PR #54, opened 2020-05-19
Federate

mew2ub:
The Federate class extends App and manages a connection to the RTI. A federate can have timed and untimed federate ports which can exchange messages with other federates through the RTI.

- Normally I would have put the Federate specific code into the Federate class and overridden functions from the App base class that needed to be changed. But those functions (like _next) are private and I know there was a lot of effort put into making sure user-defined apps and reactors created by extending the App and Reactor classes wouldn't have access to private variables and functions. So I modified the base class functions with guards like

```
if (this instanceof Federate) {
    ...
}
```

for Federate specific code. The downside to doing it this way is these checks will slow down execution a little in non-federate Apps, and the base class has higher coupling with the subclass. The advantage is private functions stayed private.

- I refactored most of the the _start() function in the App class into a new function called _init() in the same class. I did this because federate._start() has to separate initialization and scheduling startup because of its interaction with the RTI. There should be no change in functionality if you call app._start() because _start() calls _init(). 

- I added support to the time.ts module for converting TimeValues into 64 bit binary buffers. You can create a TimeValue from a 64bit binary buffer using the new BinaryTimeValue class and get a 64 bit binary buffer using the get64Bit() function on a TimeValue. The C RTI expects this kind of time representation so this makes communication with the C RTI possible.

- I prepared (and tested against) a couple simple examples. reactor-ts/src/example contains four programs: FederateTimedDest, FederateDest, FederateTimedSender, and FederateSender. These examples are meant to immitate the C target example for the RTI at lingua-franca/example/Distributed.  FederateTimedDest and FederateDest demonstrate a Federate with a TimedFederateInPort and a FederateInPort respectively. FederateTimedSender and FederateSender demonstrate a Federate with a TimedFederateOutPort and a FederateOutPort respectively. The Sender programs deliberately send a long message to test out message chunking. To run the example programs:

1) Build the reactor-ts repo via $npm run build.
2) Compile the C RTI example program at lingua-franca/example/Distributed.
3) Start the generated C RTI program at lingua-franca/example/Distributed/bin/Distributed_RTI

Note that Distributed_RTI is generated with information about this specific Federate topology, and won't work with any other model than a Sender and a Destination with the correct PortIDs and FederateIDs.

4) Start the example TS programs at reactor-ts/dist/example/FederateTimedDest.js and reactor-ts/dist/example/FederateTimedDest.js  (or the untimed versions)

You can also try exchanging one of the TS examples with its corresponding C version lingua-franca/example/Distributed/bin/Distributed_Receiver or lingua-franca/example/Distributed/bin/Distributed_Sender.


lhstrh:
I have started to look at these changes to see how we can address the numerous merge conflicts. Why are objects of type `FederateInPort` used as an argument to `schedule`? Ports aren't schedulable entities in the reactor model; only actions are...

mew2ub:
It might be best to hold off looking at the Federate branch for now because
I had a conversation with Edward about how Federates work in C, and there
are incompatibilities. I'm going to have to make some significant changes
as I figure out how to address them.

lhstrh:
Well, maybe we should also have a conversation then to avoid those changes from causing incompatibilities on the reactor-ts end. I think it would help smooth the process of integration...

mew2ub:
I'm going to have to spend some time messing around to figure out what's
going on. A conversation sounds good once I have a better sense of what I'm
doing.

I'm aware there are issues with FederateInPorts. That's one of the things
I'm fundamentally reworking.

edwardalee:
Part of the tension here is that the C code generator is primarily relying on AST transformation to handle networked communication. The connection between a port in one federate and a port in another gets broken (at the AST level), and replaced with a pair of reactions in the top-level reactor, one that reacts to the output by launching the message in the network on the sending end. On the receiving end, there is a thread that is listening for incoming messages on the socket. When it gets one, it schedules an action that triggers the second reaction to send the message to the receiving port.

I suggest that reactor.ts, instead of handling federations at such a fundamental level, just provide the library functions to make it easy for a person to write these two reactions and start the socket listener.

Note that the AST transformation is creating reactions, not reactors, which as Christian's bug identified, is problematic.  We need to either replace these reactions with reactors, as Marten did for "after", or provide a way to specify "unordered" reactions.  I'm increasingly thinking that "unordered" reactions may be useful anyway, more broadly, and is probably quite easy to implement. Should we reopen that issue?

lhstrh:
@mew2ub: I think the solution we describe in the PLDI/FDL paper applies directly: we just need two library reactors, an `RTISender` and an `RTIReceiver`. Why not just do it this way?

@edwardalee: Unordered reactions we actually already describe in the CyPhy paper (see summary in the appendix), and we primarily intended them for reactions that convey values between ports across layers of hierarchy (a notational alternative for the hierarchical connections we allow in LF). Needless to say, I agree they could be useful. The reason I've been hesitant about exposing them in the language is that it may prove difficult for programmers to judge whether it is "safe" to mark a reaction as unordered.

mew2ub:
Notes on updates to federate branch:

These changes give Reactor.ts support for federated execution.

The FederatedApp class extends App and manages a connection to the RTI on behalf of the federate (top level reactor) assigned to it. Establishing physical/logical connections between federates is done by creating physical/logical actions. A message is sent by invoking the `util.sendRTITimedMessage` from within a reaction. A message is received by registering the connection's action via the `this.registerFederatePortAction` function in the FederatedApp's constructor. 

- Normally I would have put the Federate specific code into the Federate class and overridden functions from the App base class that needed to be changed. But those functions (like _next) are private and I know there was a lot of effort put into making sure user-defined apps and reactors created by extending the App and Reactor classes wouldn't have access to private variables and functions. So I modified the base class functions with guards like

```
if (this instanceof FederatedApp) {
    ...
}
```

for FederatedApp specific code. The downside to doing it this way is these checks will slow down execution a little in non-federate Apps, and the base class has higher coupling with the subclass. The advantage is private functions stayed private.

- I refactored the _start() function in the App class into two new functions: `_checkPrecedenceGraph()` and `_alignStartAndEndOfExecution()`. I did this because federatedApp._start() has to separate initialization and scheduling startup because of its interaction with the RTI. There should be no change in functionality if you call app._start().

- I added support to the time.ts module for converting TimeValues into 64 bit binary buffers. You can create a TimeValue from a 64bit binary buffer using the new BinaryTimeValue class and get a 64 bit binary buffer using the get64Bit() function on a TimeValue. The C RTI expects this kind of time representation so this makes communication with the C RTI possible.

- I prepared (and tested against) a couple simple examples. reactor-ts/src/example contains two directories DistributedPhysical and DistributedLogical, each with a an example sender and receiver program for a logical and a physical connection. These examples are meant to immitate the C target example for the RTI at lingua-franca/example/Distributed. To run the example programs

1) Build the reactor-ts repo via $npm run build.
2) Compile the C RTI example program at lingua-franca/example/Distributed.
3) Start the generated C RTI program at lingua-franca/example/Distributed/bin/Distributed_RTI

Note that Distributed_RTI is generated with information about this specific Federate topology, and won't work with any other model than a Sender and a Destination with the correct PortIDs and FederateIDs.

4) You can also try substituting one of the TS examples with its corresponding C version lingua-franca/example/Distributed/bin/Distributed_Receiver or lingua-franca/example/Distributed/bin/Distributed_Sender.

PR #55, opened 2020-06-25
Federate

mew2ub:


mew2ub:
Notes on updates to federate branch:

These changes give Reactor.ts support for federated execution.

The FederatedApp class extends App and manages a connection to the RTI on behalf of the federate (top level reactor) assigned to it. Establishing physical/logical connections between federates is done by creating physical/logical actions. A message is sent by invoking the `util.sendRTITimedMessage` from within a reaction. A message is received by registering the connection's action via the `this.registerFederatePortAction` function in the FederatedApp's constructor. 

- Normally I would have put the Federate specific code into the Federate class and overridden functions from the App base class that needed to be changed. But those functions (like _next) are private and I know there was a lot of effort put into making sure user-defined apps and reactors created by extending the App and Reactor classes wouldn't have access to private variables and functions. So I modified the base class functions with guards like

```
if (this instanceof FederatedApp) {
    ...
}
```

for FederatedApp specific code. The downside to doing it this way is these checks will slow down execution a little in non-federate Apps, and the base class has higher coupling with the subclass. The advantage is private functions stayed private.

- I refactored the _start() function in the App class into two new functions: `_checkPrecedenceGraph()` and `_alignStartAndEndOfExecution()`. I did this because federatedApp._start() has to separate initialization and scheduling startup because of its interaction with the RTI. There should be no change in functionality if you call app._start().

- I added support to the time.ts module for converting TimeValues into 64 bit binary buffers. You can create a TimeValue from a 64bit binary buffer using the new BinaryTimeValue class and get a 64 bit binary buffer using the get64Bit() function on a TimeValue. The C RTI expects this kind of time representation so this makes communication with the C RTI possible.

- I prepared (and tested against) a couple simple examples. reactor-ts/src/example contains two directories DistributedPhysical and DistributedLogical, each with a an example sender and receiver program for a logical and a physical connection. These examples are meant to immitate the C target example for the RTI at lingua-franca/example/Distributed. To run the example programs

1) Build the reactor-ts repo via $npm run build.
2) Compile the C RTI example program at lingua-franca/example/Distributed.
3) Start the generated C RTI program at lingua-franca/example/Distributed/bin/Distributed_RTI

Note that Distributed_RTI is generated with information about this specific Federate topology, and won't work with any other model than a Sender and a Destination with the correct PortIDs and FederateIDs.

4) You can also try substituting one of the TS examples with its corresponding C version lingua-franca/example/Distributed/bin/Distributed_Receiver or lingua-franca/example/Distributed/bin/Distributed_Sender.

src/core/time.ts
```diff
@@ -212,6 +212,48 @@ export class TimeValue {
             }
         } 
     }
+
```
lhstrh, line 215:
```suggestion
public static fromBinary(buffer: Buffer): TimeValue {
        const billion = BigInt(TimeUnit.secs);

        // To avoid overflow and floating point errors, work with BigInts.
        let bigTime = buffer.readBigUInt64LE(0);
        let bigSeconds = bigTime / billion;
        let bigNSeconds = bigTime % billion;

        return new TimeValue(Number(bigSeconds), Number(bigNSeconds));
}
```

src/core/time.ts
```diff
@@ -212,6 +212,48 @@ export class TimeValue {
...
+    constructor(buffer: Buffer) {
+        const billion = BigInt(TimeUnit.secs);
+
+        // To avoid overflow and floating point errors, work with BigInts.
+        let bigTime = buffer.readBigUInt64LE(0);
+        let bigSeconds = bigTime / billion;
+        let bigNSeconds = bigTime % billion;
+
+        super(Number(bigSeconds), Number(bigNSeconds));
+    }
```
lhstrh, line 256: There's really no need to subclass `TimeValue` for a constructor call with some precomputed arguments. It's better to use a static "factory" method for that. I provided a code suggestion for this.

src/core/time.ts
```diff
@@ -212,6 +212,48 @@ export class TimeValue {
             }
         } 
     }
+
+    /**
+     * Get a 64 bit binary, little endian representation of this TimeValue.
+     * Used by federates.
+     */
+    public get64Bit() : Buffer {
```
lhstrh, line 220:
```suggestion
    public toBinary() : Buffer {
```

src/example/DistributedLogical/Distributed_msg.ts
```diff
@@ -0,0 +1,143 @@
+// Code generated by the Lingua Franca compiler from file:
+// /Users/matt.weber/git/lingua-franca/example/DistributedTS/Distributed.lf
```
lhstrh, line 2:
```suggestion
// https://github.com/icyphy/lingua-franca/blob/master/example/Distributed/Distributed.lf
```

src/example/DistributedPhysical/Distributed_dsp.ts
```diff
@@ -0,0 +1,142 @@
+// Code generated by the Lingua Franca compiler from file:
+// /Users/matt.weber/git/lingua-franca/example/DistributedTS/Distributed.lf
```
lhstrh, line 2:
```suggestion
// https://github.com/icyphy/lingua-franca/blob/master/example/Distributed/Distributed.lf
```

src/example/DistributedPhysical/Distributed_msg.ts
```diff
@@ -0,0 +1,142 @@
+// Code generated by the Lingua Franca compiler from file:
+// /Users/matt.weber/git/lingua-franca/example/DistributedTS/Distributed.lf
```
lhstrh, line 2:
```suggestion
// https://github.com/icyphy/lingua-franca/blob/master/example/Distributed/Distributed.lf
```

src/example/DistributedPhysical/Distributed_dsp.ts
```diff
@@ -0,0 +1,142 @@
+// Code generated by the Lingua Franca compiler from file:
```
lhstrh, line 1: Generally, we should not be checking in generated code, but I see why you'd want to have this readily available in the example directory...

src/example/DistributedLogical/Distributed_dsp.ts
```diff
@@ -0,0 +1,144 @@
+// Code generated by the Lingua Franca compiler from file:
+// /Users/matt.weber/git/lingua-franca/example/DistributedTS/Distributed.lf
```
lhstrh, line 2:
```suggestion
// https://github.com/icyphy/lingua-franca/blob/master/example/Distributed/Distributed.lf
```

src/core/reactor.ts
```diff
@@ -2025,8 +2046,47 @@ export class App extends Reactor {
...
         }
+        
+        public sendRTIMessage(data: Buffer, destFederateID: number, destPortID: number) {
+            return this.app._sendRTIMessage(data, destFederateID, destPortID);
+        };
+
+        public sendRTITimedMessage(data: Buffer, destFederateID: number, destPortID: number) {
+            return this.app._sendRTITimedMessage(data, destFederateID, destPortID);
+        };
+
```
lhstrh, line 2057: It doesn't seem to me that these functions are supposed to be called directly from reaction code. As such, they don't belong in `AppUtils`.

src/core/time.ts
mew2ub: I'm fine with this change.

src/example/DistributedPhysical/Distributed_dsp.ts
mew2ub: Yes, normally I would not check in generated code.

The examples are here because I wanted to make it possible to test modifications to the reactor-ts submodule without resolving my pull request to the LF repo. The code isn't 100% generated either. I started with generated code, but I modified these by hand to make them as simple as possible to work with.

src/example/DistributedLogical/Distributed_msg.ts
mew2ub: This could just be removed entirely. These programs started off generated, but I cleaned them up a bit.

src/example/DistributedPhysical/Distributed_dsp.ts
mew2ub: Could just be removed.

src/core/reactor.ts
mew2ub: This wasn't my first choice as an implementation, but I believe its necessary for these functions to be called directly in a reaction because of how federates are handled in LF. A syntax tree modification is made to the program (see the function makeCommunication on line 333 of ASTUtils) that removes connections between federates and replaces them with a sending reaction in the federated reactor triggered by the sending federate's outport (generateNetworkSenderBody), an action in the receiving federate, and a receiving reaction triggered by the generated action that sends the message to the receiving federate's inport (generateNetworkReceiverBody). For this scheme to work, the generateNetworkSenderBody reaction needs access to sendRTIMessage, and sendRTITimedMessage.

I don't like how making the functions available here makes it so arbitrary non-generated reactions might erroneously call these functions, but I didn't see another way to do it without seriously rearchitecting things. I'm open to better solutions.

lhstrh: I think this is a symptom of a bigger issue with the design.

The reaction code inside of a reactor should be completely agnostic of whether the reactor is part of a federated execution or not. While this may be the case for the Lingua Franca program, it is clearly not the case in the reactor-ts implementation. I don't see how this can even be used meaningfully _sans_ LF, which is explicitly one of the goals of reactor-ts.

I think that all of the extra plumbing that comes into play for federated execution should to be carried out somewhere else; not inside of the (generated) reaction body. In C, this isn't a problem, because the C runtime isn't intended to be used without the LF compiler. And, yes, making this suitable for standalone use would require rearchitecting things. Maybe it's too soon for a code review and we should do a design review first?

edwardalee: There may be a fundamental conflict here between two goals.  The first goal is to build infrastructure that works for all target languages. AST transformations are a key tool to accomplish this. The second goal is to have a framework that is independent of Lingua Franca with all the same capabilities.  These two goals are at odds with one another. You cannot accomplish both.  I personally feel that at this stage of development, the first goal is more important.

lhstrh: I'm not sure I buy this argument. Unlike other AST transformations (such as the one for `after`) we don't get functionality "for free"; a considerable amount of additional runtime support is necessary to make federation work. We're looking at ~1000 lines of code changed in reactor-ts! If we make such substantial changes then I'd like them to be well aligned with the goal of standalone use as well.

I guess @cmnrd might feel the same about `reactor-cpp`. To me, it makes more sense to come up with a solution that works more or less the same for C++ and TS rather than build C++ and TS implementations based on a runtime written in C.

mew2ub: Some thoughts:

- If you call sendRTIMessage or sendRTITimedMessage from a reaction belonging to a non-federated reactor it throws an error. Personally, I think putting two situationally valid functions in reaction scope is a pretty good compromise between the LF and reactor-ts design goals, since it makes everything work.

- If we were to swap out the sendRTIMessage or sendRTITimedMessage functions for some other mechanism, it would be a relatively minor change to this branch. 

> We're looking at ~1000 lines of code changed in reactor-ts!

- The majority of those ~1000 lines are new additions, not changes. I kept the changes to reactor-ts as minimal as possible. My first comment for the pull request outlines the changes. The vast majority of the new lines belong to a new class for interfacing with the RTI. The primary reason the new classes are not in a separate file is I needed to use a couple functions that were intentionally not exported or exposed from the App class.

lhstrh: > * If you call sendRTIMessage or sendRTITimedMessage from a reaction belonging to a non-federated reactor it throws an error. Personally, I think putting two situationally valid functions in reaction scope is a pretty good compromise between the LF and reactor-ts design goals, since it makes everything work.

I'm not particularly concerned about accessing those functions being harmful. I'm concerned about implementing support for federation that is unusable without the compiler tool chain. Suppose I want to write a federated reactor program in reactor-ts; how do I do this?

> * If we were to swap out the sendRTIMessage or sendRTITimedMessage functions for some other mechanism, it would be a relatively minor change to this branch.

That I am not convinced of. But, if you're right, we should just go ahead and make that minor change.

> * The majority of those ~1000 lines are new additions, not changes.

New additions _are_ changes. My point was that a substantial amount of extra code is required to support federation. Unlike support for `after`, which requires 0 extra code.

> I kept the changes to reactor-ts as minimal as possible. My first comment for the pull request outlines the changes. The vast majority of the new lines belong to a new class for interfacing with the RTI. The primary reason the new classes are not in a separate file is I needed to use a couple functions that were intentionally not exported or exposed from the App class.

I see. Though I really think these classes should not go into `reactor.ts` but into `rti.ts` or something like that. Let's look at this together to find a solution to the visibility issue. What, specifically, is the non-exported functionality you're talking about?

mew2ub: > Suppose I want to write a federated reactor program in reactor-ts; how do I do this?

You write a program like the examples I included in `src/example/DistributedPhysical` and `src/example/DistributedLogical`. They're hand-written examples. Basically what you have to do is implement the AST modification (i.e. sending reaction, receiving action, receiving reaction) yourself. If we so desired, there could be reactor-ts specific syntactic sugar for this. Something like a `connectFederates` function that sets this up behind the scenes.

> If we were to swap out the sendRTIMessage or sendRTITimedMessage functions for some other mechanism, it would be a relatively minor change to this branch.

>> That I am not convinced of. But, if you're right, we should just go ahead and make that minor change.

To clarify this point, `sendRTIMessage` and `sendRTITimedMessage` are necessary for compatibility with LF the way it currently is. _I don't think they should be removed_. But if we were to write syntactic sugar for this (or hypothetically remove the functions for some other abstraction) my experience working on this suggests it would be relatively straightforward. This is my second iteration of federates. If you were to look at the history of this branch (78ae3c59f3fc0cbb12d6b0ebd6cd5d3cdb2fe8b4) you'd see my first implementation of federates automatically sent a message to the RTI when a message was delivered to a `FederatePort`. It was about 150 lines with comments. Swapping FederatePorts out for what you see here meant removing that and writing an additional 40 lines (with comments).

>> I kept the changes to reactor-ts as minimal as possible. My first comment for the pull request outlines the changes. The vast majority of the new lines belong to a new class for interfacing with the RTI. The primary reason the new classes are not in a separate file is I needed to use a couple functions that were intentionally not exported or exposed from the App class.

> I see. Though I really think these classes should not go into reactor.ts but into rti.ts or something like that. Let's look at this together to find a solution to the visibility issue. What, specifically, is the non-exported functionality you're talking about?

Incidentally, a file named rti.ts was the first thing I tried.

- `sendRTIMessage` and `sendRTITimedMessage` involves a change to AppUtils. Potentially AppUtils could extended it to FederateUtils, but it's not exported. The reaction sandboxing mechanism is an obstacle to this too.
- federated execution requires a modification to the way logical time advancement is handled. `_next` is private.
- shutdown needs to do a bit of RTI communication and clean up the socket. Most of the shutdown code is private/non-exported.

lhstrh: I just checked out your branch and made changes that will allow you to simply move everything below line 2644 into a separate file. I prefer the name `federation.ts` because it is less cryptic and more descriptive of what the code facilitates than `rti.ts`. Please go ahead and do this.

All tests should pass after you update the imports of the files that use `FederatedApp` (i.e., now use
`import {FederatedApp} from '../../core/federation'`.)

edwardalee: > I'm not sure I buy this argument. Unlike other AST transformations (such as the one for `after`) we don't get functionality "for free"; a considerable amount of additional runtime support is necessary to make federation work. We're looking at ~1000 lines of code changed in reactor-ts! If we make such substantial changes then I'd like them to be well aligned with the goal of standalone use as well.
> 
> I guess @cmnrd might feel the same about `reactor-cpp`. To me, it makes more sense to come up with a solution that works more or less the same for C++ and TS rather than build C++ and TS implementations based on a runtime written in C.

This is a valid point, but will federations will be designed with unmodified library reactors? Even if they are, is wrapping the library reactors really a big deal? It will be needed anyway to deal with decentralized communication. I haven't written any reactor-ts code, so I'm probably not qualified to judge, but sendRTITimedMessage() seems like a reasonable approach to me.  I would guess someone building a distributed system using this framework will really have to be aware of the distribution architecture, and making that awareness explicit in the code doesn't seem like such a bad thing to me.

On the reactor-cpp framework, I've started to look at getting our C generated code to compile with g++. It works already with HelloWorld.lf, but more work is needed for the others. Soroush suggested trying this. It could also pave a path towards a very easy Python target. Soroush pointed out that many other polyglot systems like Tensorflow have the core written in C surrounded by wrappers for various user languages.

lhstrh: > This is a valid point, but will federations will be designed with unmodified library reactors?

I don't see why not?

> Even if they are, is wrapping the library reactors really a big deal? It will be needed anyway to deal with decentralized communication. I haven't written any reactor-ts code, so I'm probably not qualified to judge, but sendRTITimedMessage() seems like a reasonable approach to me. I would guess someone building a distributed system using this framework will really have to be aware of the distribution architecture, and making that awareness explicit in the code doesn't seem like such a bad thing to me.

I'm not making a case against wrapping library reactors. All I'm saying is that the implementation in this PR is not suitable for standalone use. It functions fine in tandem with Lingua Franca, but to "implement the AST modification by hand" and manually figure out the respective PortIDs and FederateIDs is not a reasonable task to be carried out by a programmer who wants to use reactor-ts as a framework rather than a compiler target.

Up until this PR, I was literally never asked to give my opinion about any of the design decisions that lead to this particular implementation of federation in reactor-ts, which I find unfortunate. In my experience, it is much more productive to have these kinds of discussions earlier on in the process...

> On the reactor-cpp framework, I've started to look at getting our C generated code to compile with g++. It works already with HelloWorld.lf, but more work is needed for the others. Soroush suggested trying this. It could also pave a path towards a very easy Python target. Soroush pointed out that many other polyglot systems like Tensorflow have the core written in C surrounded by wrappers for various user languages.

Hmm, sounds interesting. Let's make sure to bring @cmnrd in the loop as well.

lsk567: > 1. Build the reactor-ts repo via $npm run build.
> 2. Compile the C RTI example program at lingua-franca/example/Distributed.
> 3. Start the generated C RTI program at lingua-franca/example/Distributed/bin/Distributed_RTI
> 
> Note that Distributed_RTI is generated with information about this specific Federate topology, and won't work with any other model than a Sender and a Destination with the correct PortIDs and FederateIDs.
> 
> 1. You can also try substituting one of the TS examples with its corresponding C version lingua-franca/example/Distributed/bin/Distributed_Receiver or lingua-franca/example/Distributed/bin/Distributed_Sender.

@mew2ub I see that in a previous commit ([14d17cc](https://github.com/icyphy/reactor-ts/pull/55/commits/14d17cc5c042f96fcf953ec67b7727a438ebffd7)), you have removed a TypeScript implementation of the RTI. I am wondering if we are going to re-implement the RTI in TS or the TS target will share the C RTI (perhaps expanded to support more Federate topologies). Since the RTI can be a standalone general infrastructure, it seems to me that having a well-built implementation in C would be sufficient. I am not sure if this is the plan for the TS target.

lsk567:
> 1. Build the reactor-ts repo via $npm run build.
> 2. Compile the C RTI example program at lingua-franca/example/Distributed.
> 3. Start the generated C RTI program at lingua-franca/example/Distributed/bin/Distributed_RTI
> 
> Note that Distributed_RTI is generated with information about this specific Federate topology, and won't work with any other model than a Sender and a Destination with the correct PortIDs and FederateIDs.
> 
> 1. You can also try substituting one of the TS examples with its corresponding C version lingua-franca/example/Distributed/bin/Distributed_Receiver or lingua-franca/example/Distributed/bin/Distributed_Sender.

@mew2ub I see that in a previous commit ([14d17cc](https://github.com/icyphy/reactor-ts/pull/55/commits/14d17cc5c042f96fcf953ec67b7727a438ebffd7)), you have removed a TypeScript implementation of the RTI. I am wondering if we are going to re-implement the RTI in TS or the TS target will share the C RTI (perhaps expanded to support more Federate topologies). Since the RTI can be a standalone general infrastructure, it seems to me that having a well-built implementation in C would be sufficient. I am not sure if this is the plan for the TS target.

src/core/reactor.ts
```diff
@@ -3192,6 +3184,49 @@ export class FederatedApp extends App {
...
+     * If this federated app has not received a sufficiently large time advance
+     * grant (TAG) from the RTI for the next event, send it a Next Event Time
+     * (NET) message and return. _next() will be called when a new greatest TAG
+     * is received. The NET message is not sent if the connection to the RTI is
+     * closed. FIXME: what happens in that case? Will next be called?
+     * @param event 
+     */
+    protected canProceed(event: TaggedEvent<Present>) {
+        if (this._isRTISynchronized() && this._isActive()) {
+            // FIXME: Why would it proceed if the reaction is inactive?
```
mew2ub, line 3207: It looks like the check for `this._isActive()` was a bandaid I added here when I was hunting down a bigger bug and isn't necessary. I'll remove it.

mew2ub:
@lsk567 Yeah that's exactly right. I started off implementing the RTI in TypeScript, but then I had a conversation with Edward about it and we agreed TypeScript federates could work with the existing C RTI. So, I believe we're planning to use the C RTI for all targets right now. Perhaps in the very long term we might want to reimplement the RTI in a different language, but there's no urgent need to do that.

src/core/reactor.ts
lhstrh: Shouldn't `canProceed` return immediately if the reactor is not active, without contacting the RTI?

src/core/time.ts
lhstrh: OK, can you remove this class and change all references accordingly?

src/core/time.ts
lhstrh: This is just for symmetry in the naming (i.e., `fromBinary` and `toBinary`). If you could do the rename that would be great. It can also be `to64Bit` and `from64Bit` -- up to you.

src/example/DistributedLogical/Distributed_msg.ts
lhstrh: We should not remove it but instead point to a valid LF source, as I proposed. Without that reference the example suggests people are to write federated reactors by hand, which I certainly wouldn't recommend.

src/example/DistributedPhysical/Distributed_dsp.ts
lhstrh: I actually think these examples are best put in the `lingua-franca` repo, not the `reactor-ts` repo. In that case we also don't have to check in generated code, which is generally considered a bad practice.

This also makes it abundantly clear that the expected workflow for writing federated reactors is to use the compiler tool chain.

In short, I think that examples in reactor-ts should feature hand-written reactors only.

src/core/reactor.ts
mew2ub: In your suggestion, how should it return? True or false? My understanding of the purpose of TAGs in the RTI is that logical time in a federate should only be allowed to advance up to the greatest received TAG. Why should an inactive FederatedApp get special treatment?

src/example/DistributedPhysical/Distributed_dsp.ts
mew2ub: I have a different example program checked into the LF repo (which is a .lf file), so I don't think we should move this there.

Although these examples began life as generated code, they have been significantly modified from that by hand. To me, this qualifies them as hand-written. If I were to further clean them up so they didn't have any of the code-generated boilerplate at all, would that make them check-in-able to you?

> This also makes it abundantly clear that the expected workflow for writing federated reactors is to use the compiler tool chain.

Why is this the expected workflow? It doesn't seem particularly more difficult to me to write a federated reactor by hand in TypeScript than it is to write a moderately complex non-federated reactor by hand in TypeScript. There are something like 5 extra functions you have to use to set it up.

lhstrh: I understand that generated code can be polished up to look like it was handwritten, but that doesn't make them handwritten. If this is a different example than the one in the LF repo, then why not just have two examples in the LF repo?

Having to figure out numerical federate IDs and port IDs in order to plug things together strikes me as brittle and difficult to write an API manual for. This API was written for a code generator, not for a programmer. I think we should write an API that is accessible for programmers, and until we do, I suggest not advertising this as a feature of reactor-ts. Like I stated before, this approach works fine in LF, because there all the marshaling between reactors is abstracted using ordinary connections. This is why I say using the LF compiler is the expected workflow.

src/core/reactor.ts
lhstrh: You're right. This is not the right place to do this check anyway. You can removed it.

src/core/reactor.ts
mew2ub: Something just occurred to me that was perhaps obvious to the rest of you in this discussion, but I don't think was ever explicitly stated: the biggest obstacle to handwriting federated reactors in TypeScript at the present moment is dependence on the C RTI. Since the only way to create an RTI program right now is to generate it from LF, reactor-ts can't independently support federates, period. To fully handwrite a federation you'd have to be able to handwrite the RTI's representation of the federate topology. If we decided this was a worthwhile goal, perhaps we could try to make the C RTI more accessible for handwriting or we could try to resurrect my preliminary attempt for a TypeScript RTI.

> I'm not making a case against wrapping library reactors. All I'm saying is that the implementation in this PR is not suitable for standalone use. It functions fine in tandem with Lingua Franca, but to "implement the AST modification by hand" and manually figure out the respective PortIDs and FederateIDs is not a reasonable task to be carried out by a programmer who wants to use reactor-ts as a framework rather than a compiler target.

I don't understand what you're getting at when you say this task is unreasonable. What's so bad about expecting a programmer to create reactions for network communication, and to assign numbers to their federates? Wouldn't a programmer be writing reactions for network communication if they wanted interaction with a non-federated remote service? I see this as pretty much the same thing. We could even add syntactic sugar to make it easier.

src/core/time.ts
mew2ub: Just did it.

src/core/reactor.ts
lhstrh: To be concrete, I would propose to do the following:
1. remove the generated code from this PR;
2. merge `federation.ts` and changes to `reactor.ts` into master "as is"; and
3. jointly design a new reactor-ts API for building federations _sans LF_.

As Edward suggested, 3 might not have the highest priority right now. If you want to work on this, however, let's meet to hammer out a design that we can agree on before we move to implementing it.

src/example/DistributedPhysical/Distributed_dsp.ts
mew2ub: > I understand that generated code can be polished up to look like it was handwritten, but that doesn't make it handwritten. If this is a different example than the one in the LF repo, then why not just have two examples in the LF repo?

Because these examples illustrate how you can write a federated program in TypeScript. They make it possible to try out changes to how federates work in TypeScript without having to rebuild the LF compiler. Would it satisfy your desire to not advertise these as a feature of reactorTS if I added a README into their directory, explaining that point?

> Having to figure out numerical federate IDs and port IDs in order to plug things together strikes me as brittle and difficult to write an API manual for.

To me this task is pretty similar to working with threadIDs or processIDs. Would an API manual really be that bad? "Assign each federatedApp an ID in its constructor. The ID must be unique among all federates and be a number between 0 and NUMBER_OF_FEDERATES - 1"

mew2ub:
Cool!

I still think the files you're referring to are useful and worth checking in. I counted and there are 159 lines of differences between the logical dsp example and what you get out of the code generator for the Distributed.lf. It doesn't make sense to me to remove such a modified file from the repo because it's "generated". I'm open to moving them to a different directory or to rename them or to add a README.

lhstrh:
OK, how about you move them into `example/generated` and supply a `README.md` that explains the purpose of these files and how to use them.

mew2ub:
Ah right. Incidentally, the right file to refer to is 

`https://github.com/icyphy/lingua-franca/blob/master/example/DistributedTS/Distributed.lf`

instead of

`https://github.com/icyphy/lingua-franca/blob/master/example/Distributed/Distributed.lf`

with the caveat that DistributedTS hasn't been merged into master yet because it's part of the other pull request.

PR #61, opened 2021-07-15
Fix critical TypeScript errors in federation.ts and make TypeScript federated example run at a bare minimum.

hokeun:
This fix allows the dormant TypeScript federated example to run at a bare minimum.

The example federates can now connect to the C-compiled RTI (HelloWorld_RTI) and advance time correctly. Also, now, the source reactor (in Distributed_msg.ts) can generate messages correctly, and the destination reactor (in Distributed_dsp.ts) can display messages.

Major errors fixed are:
* Message format of MSG_TYPE_FED_IDS.
* Formats of messages with missing microsteps.
* Missing MSG_TYPE_UDP_PORT message.

### DIFF
--- a/src/core/federation.ts
+++ b/src/core/federation.ts
@@ -38,15 +38,32 @@ export const CONNECT_NUM_RETRIES: number = 500;
  */
 enum RTIMessageTypes {
 
-    /**
-     * Byte Identifying a federate ID message, which is 32 bits long.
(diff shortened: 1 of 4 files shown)


lhstrh (approved):
🚀🚀🚀 Way to go! This will give us a great starting point to start fixing the code generator and adding tests.

PR #70, opened 2021-12-21
Add more tests to improve coverage of core/time.ts and core/component.ts

CloverCho:
[core/time.ts]   :  Improve line coverage from 86.9% to 100%
1. Add tests for using static methods which create TimeValue object : secs, msec, msecs, usec, usecs, nsec, nsecs
2. Add tests for using toBinary/fromBinary methods in TimeValue class
3. Add test cases for Alarm class
    a) A case that a delay to enter "set" method satisfies 0 < delay < 25ms
    b) A case that an Alarm object is already active and Immediate reference exists when "set" method is called with the object.

[core/component.ts]  :  Improve line coverage from 95.24% to 100%
1. Add tests for using methods that are missed : _getAlias, _isContainedBy, _isContainedByContainerOf


__tests__/alarm.ts
```diff
@@ -316,9 +316,76 @@ describe('Timer tests', () => {
...
+                    avgError += errors[i];
+                }
+                avgError = avgError / numSamples;
+                console.log('\t\t - Min. Error: ' + minError + '%');
+                console.log('\t\t - Max. Error: ' + maxError + '%');
+                console.log('\t\t - Avg. Error: ' + avgError + '%');
+                done();
+            } else {
+                alarm.set(task, delay10Msec, schedule);        
+            }
```
hokeun, line 381: Could we factor this out as a separate function? It looks like this part of the code is repeated multiple times.

hokeun (approved):
Nicely done! It's a good improvement. Left a minor comment.

__tests__/alarm.ts
CloverCho: I try to separate that function out but failed.
I think it is difficult because of the following reason:
The schedule function uses external variables(doneCount, numSamples, etc..), so one way to separate the function is
by replacing these variables with additional arguments.
(For example, **schedule(tv: TimeValue, doneCount: Number, numSamples: Number)** )
However, the schedule function should have a TimeValue argument only, according to line 576 of core/time.ts.
So additional arguments are invalid.

I think there are other ways to separate the function, but couldn't find it.
If I find the way, I will apply it.

hokeun: Ah, I see. No worries! Then I think it's fine to leave it as is now. Thanks for checking!

hokeun:
@lhstrh @Soroosh129  We've been getting the following error while the CI is running test-deploy:
`remote: Permission to lf-lang/reactor-ts.git denied to github-actions[bot].`
Would you know if this is relevant to the GitHub actions setup?
Details about this failure can be found here: https://github.com/lf-lang/reactor-ts/runs/4638953749?check_suite_focus=true

lhstrh:
Deployment should not be triggered by pull requests, only when pushed to master. My guess is that since @CloverCho didn't have write access, this faulty type of deployment yielded an access control error. I've given @CloverCho write access but will also remove the pull request trigger from the documentation generation/deployment task.

lhstrh (approved):
Thanks, @CloverCho! I'll go ahead and merge this in. I've given you write access, so for future changes it would be ideal if you were to create a branch in this repo (rather than your own) so we can easily collaborate.

__tests__/alarm.ts
CloverCho: You're welcome! Thank you for your attention to this code!

CloverCho:
You're Welcome! I just confirmed that I have received your invitation message. Thanks for involving me in this project!

PR #71, opened 2021-12-29
Fix federated TS behavior mismatch with federated C, using intended tag from RTI for scheduling federate port action (LF issue #647)

hokeun:
Issue: https://github.com/lf-lang/lingua-franca/issues/648

Fix federated TS behavior mismatch with federated C and add the first step for PTAG handling.
Right now, the PTAG handling only does reading the PTAG message correctly.

Actual semantic handling of PTAG will be implemented in the next PR to keep each PR as short as possible.

Also, please let me know if you have any suggestions for a better way not to add extra microstep for network messages.


lhstrh:
> Thanks for the suggestion. I added this check in [b99781d](https://github.com/lf-lang/reactor-ts/commit/b99781d0d53bf320682ab94e617e6fb9ba1cd519).

This fix will work for centralized coordination (where it is a hard error), but not for decentralized coordination, where it should lead to the invocation of a user-defined fault handler. We can create an issue for this and leave it out of this PR.

src/core/reactor.ts
```diff
@@ -354,10 +354,24 @@ export class Action<T extends Present> extends ScheduledTrigger<T> implements Re
                     tag = tag.getMicroStepLater();
                 }
             }
-    
-            if (this.action.origin == Origin.logical && !(this.action instanceof Startup)) {
+
+            if (this.action.origin == Origin.logical && !(this.action instanceof Startup)
```
lhstrh, line 358: I would suggest to put this in the `else` clause of the ` if (this.action instanceof FederatePortAction) {` below (and remove the last condition from the branch predicate).

lhstrh (approved):
These changes look good. As a general comment, `federation.ts` could use some refactoring as some of the existing functions in there are pretty huge.

src/core/reactor.ts
hokeun: Good point. Done in https://github.com/lf-lang/reactor-ts/pull/71/commits/61498a6a4afec457cffff3a550f396856a22468a.

lhstrh (approved):

hokeun:
> > Thanks for the suggestion. I added this check in [b99781d](https://github.com/lf-lang/reactor-ts/commit/b99781d0d53bf320682ab94e617e6fb9ba1cd519).
> 
> This fix will work for centralized coordination (where it is a hard error), but not for decentralized coordination, where it should lead to the invocation of a user-defined fault handler. We can create an issue for this and leave it out of this PR.

Sounds good. I created a new issue here: https://github.com/lf-lang/reactor-ts/issues/75

Soroosh129:
I don't mean to nitpick, but the name of tag field carried on timed messages is `intended tag`, not indicated tag.

hokeun:
> I don't mean to nitpick, but the name of the tag field carried on timed messages is `intended tag`, not indicated tag.

lol not sure what happened in the beginning. Fixed the terms. Thanks for catching this.

PR #73, opened 2022-01-03
Perform integration tests as part of CI

lhstrh:


.github/workflows/extract-ref.yml
```diff
@@ -0,0 +1,25 @@
+name: Extract ref from file
+
+on:
+  workflow_call:
+    inputs:
+      name:
+        description: Name of the sha1 file to extract the ref from
+        type: string
+        required: false
```
hokeun, line 9: Although I'm not sure, should this be true? It looks like this job run: requires this file name.

hokeun (approved):

.github/workflows/extract-ref.yml
lhstrh: Yes, it is required. Without the name, we don't know from which file to extract the ref.

Issue #57, opened 2021-01-14
Add support for multiports (and banks?)


lhstrh:

hokeun:
Multiport implementation is required to implement some of the benchmarks from Savina suite (e.g., Online Facility Location and UCT(Unbalanced Cobwebbed Tree)) to support tree-like actor structure, sending and receiving messages to/from multiple children.

Banks are required to implement UCT as the number of children can be arbitrary, while the number of children in Online Facility Location is always four (Quadrant actor).

hokeun:
Here are rough ideas for implementing multiports in TypeScript.

### For reactor-ts runtime support:

First we need something similar to [lfutil::bind_multiple_ports()](https://github.com/lf-lang/lingua-franca/blob/826a50888d7bfafd5223280fce8285e626ebbc39/org.lflang/src/lib/cpp/lfutil.hh#L64) in Cpp target. This function takes a vector of source ports and a vector of destination ports, and then calls bind_to() function for each individual connection.

Similarly, we can create a new function in reactor.ts, called *_connectMultiplePorts()*, which takes an Array of src ports and an Array of dst ports, and then calls _[connect()](https://github.com/lf-lang/reactor-ts/blob/b3bafb4da7c98e66ef4aec965e5c21e7d5dc012d/src/core/reactor.ts#L1391) function  internally for the given Arrays of ports.

### For code generation:

In TSPortGenerator.kt, generate multi ports as new Arrays.

In TSConnectionGenerator.kt, generate connection code using *_connectMultiplePorts()*.

hokeun:
### Note from [Edward's comments](https://github.com/lf-lang/lingua-franca/pull/942#pullrequestreview-879026731)

- `interleaved` must be taken care of when supporting banks together with multiports.

PR #81, opened 2022-02-06
Add support for connecting multiports in TypeScript runtime.

hokeun:
Relevant PR: https://github.com/lf-lang/lingua-franca/pull/942
Relevant Issue: https://github.com/lf-lang/reactor-ts/issues/57


src/core/bank.ts
```diff
@@ -25,17 +26,33 @@ type ReactorConstructor<T extends Reactor, S> = {
...
+            throw new Error("Can only establish link to runtime once. Name: " + this._getFullyQualifiedName())
+        }
+    }
 	private readonly members: Array<T> = new Array();
-	constructor(width: number, cls:ReactorConstructor<T,S>, ...args:ReactorArgs<S>) {
+	constructor(container: Reactor, width: number, cls:ReactorConstructor<T, S>, bankIndexArgIndex?: number, ...args:ReactorArgs<S>) {
+        super(container)
 		
 		for (let i=0; i < width; i++) {
+            if (bankIndexArgIndex !== undefined) {
```
lhstrh, line 44: This seems rather hacky and would be easy to get wrong when you're instantiating a bank manually...

__tests__/ActionTrigger.test.ts
```diff
@@ -1,6 +1,4 @@
-import {App, Triggers, Args} from '../src/core/reactor';
-import {Origin, TimeValue} from '../src/core/time';
-import {Reactor, Timer, Action, Sched} from '../src/core/reactor';
+import {App, Triggers, Args, Origin, TimeValue, Reactor, Timer, Sched, Action} from '../src/core/internal';
```
lhstrh, line 1: Tests should probably not use the internal exports but the external ones, which we yet have to set up.

src/core/types.ts
```diff
@@ -0,0 +1,109 @@
+
```
lhstrh, line 1: Note sure whether we should have a dedicated file for types. Probably better to put these in files together with classes they relate to.

lhstrh (approved):
There a lot that can still be improved upon, but this is already a good step in the right direction. Perhaps we should prioritize more cleanups and getting `reactor-ts` published on NPM so we can have a cleaner integration with LF (rather than copy code from a submodule).

PR #89, opened 2022-03-09
Add support for ports from banks.

hokeun:
Extend Variable and Triggers to array of Variables and array of Triggers, respectively.
Add allWritable() and writable() for Bank, as the ports from banks can be used as effects

### DIFF
--- a/src/core/types.ts
+++ b/src/core/types.ts
@@ -8,7 +8,7 @@ import { Tag, TimeValue, Trigger } from "./internal";
  * @see Write
  * @see Sched
  */
- export type Variable = Read<unknown> | MultiRead<unknown>
+ export type Variable = Read<unknown> | MultiRead<unknown> | Array<Read<unknown> | MultiRead<unknown>>
 
 /**
  * Interface for writable ports.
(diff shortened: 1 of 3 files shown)


src/core/types.ts
```diff
@@ -92,8 +92,8 @@ export class Args<T extends Variable[]> {
 }
 
 export class Triggers {
-    list: Trigger[];
-    constructor(trigger: Trigger, ...triggers: Trigger[]) {
+    list: Array<Trigger | Array<Trigger>>;
+    constructor(trigger: Trigger | Array<Trigger>, ...triggers: Array<Trigger | Array<Trigger>>) {
```
lhstrh, line 96: With this change, I think it's also necessary to handle the `Array<Trigger>` case in `addReaction`/`_recordDeps` since we those not match the `instanceof Trigger` type guards.

hokeun: Good points, thanks! I think I addressed your points here: https://github.com/lf-lang/reactor-ts/pull/89/commits/8c6d4b52f5b837452858a21adc7cf254d35fdf09

Please take another look!

lhstrh: Looks good!

lhstrh (approved):

PR #90, opened 2022-03-22
Improved test coverage and fixed conversion of tags to binary strings

mattchorlian:


src/core/time.ts
```diff
@@ -431,7 +431,7 @@ export class Tag {
         let timeBuffer = this.time.toBinary();
         let buf = Buffer.alloc(12);
         timeBuffer.copy(buf, 0, 0);
-        buf.writeInt32LE(this.microstep);
+        buf.writeInt32LE(this.microstep, 8);
```
lhstrh, line 434: This is why we need unit tests! Thanks for catching this :-) 🚀🚀🚀

lhstrh (approved):

PR #91, opened 2022-03-27
Fix issues in federated execution of TypeScript.

hokeun:
1. Fix bug in LTC message
2. Handle PTAG properly
3. Assume PTAG received at the beginning for a federate.
4. Fix ow util.requestStop() is handled.

Relevant PR: https://github.com/lf-lang/lingua-franca/pull/1062

### DIFF
--- a/src/core/action.ts
+++ b/src/core/action.ts
@@ -83,11 +83,11 @@ export abstract class SchedulableAction<T extends Present> implements Sched<T> {
                 if (intendedTag === undefined) {
                     throw new Error("FederatedPortAction must have an intended tag from RTI.");
                 }
-                if (!this.action.runtime.util.isLastTagProvisional() && intendedTag <= this.action.runtime.util.getCurrentTag()) {
+                if (!this.action.runtime.util.isLastTAGProvisional() && intendedTag <= this.action.runtime.util.getCurrentTag()) {
                     throw new Error("Intended tag must be greater than current tag. Intended tag" +
                         intendedTag + " Current tag: " + this.action.runtime.util.getCurrentTag());
                 }
-                if (this.action.runtime.util.isLastTagProvisional() && intendedTag < this.action.runtime.util.getCurrentTag()) {
+                if (this.action.runtime.util.isLastTAGProvisional() && intendedTag < this.action.runtime.util.getCurrentTag()) {
                     throw new Error("Intended tag must be greater than or equal to current tag." +
                         "when the last tag is provisional Intended tag" + intendedTag +
                         " Current tag: " + this.action.runtime.util.getCurrentTag());
--- a/src/core/action.ts
+++ b/src/core/action.ts
@@ -83,9 +83,14 @@ export abstract class SchedulableAction<T extends Present> implements Sched<T> {
                 if (intendedTag === undefined) {
                     throw new Error("FederatedPortAction must have an intended tag from RTI.");
                 }
-                if (intendedTag <= this.action.runtime.util.getCurrentTag()) {
+                if (!this.action.runtime.util.isLastTagProvisional() && intendedTag <= this.action.runtime.util.getCurrentTag()) {
                     throw new Error("Intended tag must be greater than current tag. Intended tag" +
-                    intendedTag + " Current tag: " + this.action.runtime.util.getCurrentTag());
+                        intendedTag + " Current tag: " + this.action.runtime.util.getCurrentTag());
+                }
+                if (this.action.runtime.util.isLastTagProvisional() && intendedTag < this.action.runtime.util.getCurrentTag()) {
+                    throw new Error("Intended tag must be greater than or equal to current tag." +
+                        "when the last tag is provisional Intended tag" + intendedTag +
+                        " Current tag: " + this.action.runtime.util.getCurrentTag());
                 }
                 Log.debug(this, () => "Using intended tag from RTI, similar to schedule_at_tag(tag) with an intended tag: " +
                 intendedTag);
--- a/src/core/federation.ts
+++ b/src/core/federation.ts
@@ -840,7 +840,7 @@ export class FederatedApp extends App {
         let currentTime = this.util.getCurrentLogicalTime()
         if (currentTime.isEarlierThan(nextTag.time)) {
             // Tell the RTI logical time is being advanced to a greater value.
-            this.sendRTILogicalTimeComplete(currentTime);
+            this.sendRTILogicalTimeComplete(nextTag.time);
         }
         super._advanceTime(nextTag)
     }
@@ -963,8 +963,8 @@ export class FederatedApp extends App {
      * advance logical time.
      */
     public sendRTINextEventTime(nextTag: Tag) {
+        Log.debug(this, () => {return `Sending RTI next event time with time: ${nextTag}`});
         let tag = nextTag.toBinary();
-        Log.debug(this, () => {return `Sending RTI next event time with time: ${tag}`});
         this.rtiClient.sendRTINextEventTime(tag);
     }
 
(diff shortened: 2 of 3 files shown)


src/core/reactor.ts
```diff
@@ -1608,6 +1609,13 @@ export class App extends Reactor {
      */
     private _reactorsToRemove = new Array<Reactor>();
 
+    /**
+     * Stores whether the last received Tag was provisional.
+     * Every federate starts out assuming that it has been granted a PTAG
+     * at the start time, or if it has no upstream federates, a TAG.
+     */
+    protected _isLastTagProvisional: boolean = false;
```
hokeun, line 1617: Rename this to isCurrentTagProvisional (also suggest change in reactor-c).

Soroosh129: The field is `is_last_TAG_provisional` in reactor-c. Notice the capitalized TAG, which stands for Tag Advance Grant. This field indicates whether or not the last TAG message received from the RTI was a PTAG. This field is not related to the current tag.

src/core/federation.ts
```diff
@@ -839,17 +840,16 @@ export class FederatedApp extends App {
         let currentTime = this.util.getCurrentLogicalTime()
         if (currentTime.isEarlierThan(nextTag.time)) {
             // Tell the RTI logical time is being advanced to a greater value.
-            this.sendRTILogicalTimeComplete(currentTime);
+            this.sendRTILogicalTimeComplete(nextTag.time);
```
hokeun, line 843: Check c implementation and see when the LTC (Logical Time Complete) should be sent. (Probably it should be sent after processing everything in the event queue.) And make the naming make sense.

src/core/reactor.ts
lhstrh: This is super confusing. I understand why @hokeun would write `_isLastTagProvisional` instead of the correct spelling `_isLastTAGProvisional` which is only marginally different. The adoption of `TAG` as a message type is really unfortunate... Why do we need the `T`? Isn't "Advance Grant" clear enough? Are there any other things granted? Regardless, a less error-prone name for the field could be `_isLastGrantProvisional`.

src/core/federation.ts
hokeun: See what MSG_TYPE_LOGICAL_TAG_COMPLETE message is for.

https://github.com/lf-lang/reactor-c/blob/main/core/federated/net_common.h#L403

src/core/reactor.ts
Soroosh129: TAG is a recursive acronym for Tag Advance Grant :) 
It is a descendant of `TimeAdvancedGrant` in HLA.

With that said, I don't personally have a problem with renaming it.
I think `advance grant` is a bit vague since it is not clear what is meant to advance.

Nonetheless, I think in the context of this particular flag, the documentation is very elaborate:
```C
/**
 * Indicates whether the last TAG received is provisional or an ordinary
 * TAG. 
 * If the last TAG has been provisional, network control reactions must be inserted.
 * This variable should only be accessed while holding the mutex lock.
 */
```

src/core/federation.ts
Soroosh129: In reality, LTC and NET can be combined into an NMR (Next Message Request) message.
LTC is supposed to indicate to the RTI that a federate has finished processing a particular tag, usually followed by a NET that is indicating to the RTI what tag it is going to execute next, absent any unforeseen incoming network messages.

src/core/reactor.ts
lhstrh: I understand that `TAG` is a recursive acronym, which is nice, but it's also confusing because it is easily confused with the notion of a tag. For not, I think the most pragmatic thing would be to rename the field to `_isLastGrantProvisional` (and, more generally, try to avoid the term TAG in method names or fields).

Soroosh129: Maybe `_isLastRTIGrantProvisional`?

This appears to be in `reactor.ts` not `federation.ts`, which might require extra context in the name. I'm not an expert in naming conventions in TS though.

src/core/federation.ts
Soroosh129: Our reference implementation of the RTI in C expects separate LTC and NET messages, but the comment above was meant to give some context about LTC and its relation to NET.

lhstrh: Thanks for the explanation, @Soroosh129! Is this protocol described anywhere but in the comments of the C implementation? In conclusion, the used terminology makes sense, but the implementation has been wrong all along. The argument should be `currentTime` like it was, but the statement should be in a different place altogether, in a hook that executes once all reactions have executed and the reaction queue is empty.

src/core/reactor.ts
lhstrh: It's not an RTI grant though -- it's a Tag Advance Grant. Leaving off the Tag part doesn't introduce any ambiguity provided there are no other kinds of grants happening (but if there were the name you suggest would be equally ambiguous).

src/core/federation.ts
Soroosh129: Yes, this is explained in an (the) upcoming paper.

Some details about the federated execution is also explained in the "Quantifying the CAP Theorem" paper.

src/core/federation.ts
```diff
@@ -836,13 +836,12 @@ export class FederatedApp extends App {
         return true
     }
 
-    protected _advanceTime(nextTag: Tag) {
+    protected _iterationComplete(nextEvent?: TaggedEvent<Present>): void {
         let currentTime = this.util.getCurrentLogicalTime()
-        if (currentTime.isEarlierThan(nextTag.time)) {
+        if (nextEvent === undefined || currentTime.isEarlierThan(nextEvent.tag.time)) {
```
lhstrh, line 841: @Soroosh129: is a LTC expected at _any_ conclusions of a logical time? The inherited logic here only sends a `TAG` _only if_ the next tag has a greater time value (i.e., it doesn't send a TAG when only the microstep is different). Is this correct? I wasn't sure.

src/core/reactor.ts
hokeun: Thank you @Soroosh129 and @lhstrh ! I renamed this to _isLastTAGProvisional in https://github.com/lf-lang/reactor-ts/pull/91/commits/a0eb024cfe944b073cebdb739e0e52735dde3e35.
This may not be the best, but follows the naming in reactor-c. I also updated the comments.

src/core/federation.ts
Soroosh129: Microsteps are important under centralized coordination.
There should be an LTC message sent at the conclusion of each tag.

src/core/action.ts
```diff
@@ -83,9 +83,14 @@ export abstract class SchedulableAction<T extends Present> implements Sched<T> {
                 if (intendedTag === undefined) {
                     throw new Error("FederatedPortAction must have an intended tag from RTI.");
                 }
-                if (intendedTag <= this.action.runtime.util.getCurrentTag()) {
+                if (!this.action.runtime.util.isLastTagProvisional() && intendedTag <= this.action.runtime.util.getCurrentTag()) {
```
lhstrh, line 86: Should it also be `isLastTAGProvisional`?

src/core/federation.ts
lhstrh: OK, then this if-statement is wrong and should be removed. It causes `LTC` message to only be sent when there is a timevalue increment (as opposed to a microstep increment).

lhstrh (changes requested):
See comments.

src/core/action.ts
hokeun: Done!

lhstrh (approved):

PR #95, opened 2022-04-25
Make looking up bank ids easier

lhstrh:
Each reactor can look up its bank id if it has one. This PR generalizes this capability by allowing reactors to find the bank id of any of its parents by specifying a `depth` parameter in the look up method.


src/core/reactor.ts
```diff
@@ -149,11 +149,15 @@ export abstract class Reactor extends Component {
      * Return the location of the reactor instance in a bank,
```
hokeun, line 149: Maybe we should update the comments as well?

src/core/reactor.ts
```diff
@@ -1734,7 +1738,7 @@ export class App extends Reactor {
                 }
     
             } else {
-                console.log(">>>>>>>>>Postponed Scheduling of timer " + timer._getFullyQualifiedName())
+                //console.log(">>>>>>>>>Postponed Scheduling of timer " + timer._getFullyQualifiedName())
```
hokeun, line 1741: Maybe it's a good time to delete this logging, but I'm fine with leaving it as is.

hokeun (approved):
Looks good to me! Just left minor comments.

src/core/reactor.ts
lhstrh: Yep. Will do!

PR #96, opened 2022-04-25
Fix NPE in `getBankIndex`

lhstrh:


edwardalee:
Looks better, but I still think a reactor should not reference a bank index further up in the hierarchy.  See how this is done in C, where the bank index is passed down in the code that constructs the hierarchy, which is where it should be because there is where you know the depth, not in a reaction.

lhstrh:
OK, I'm fine with not having this feature, but I created it to accommodate the example you're referring to. In reactor-ts we don't pass bank indices down the hierarchy; the runtime takes care of this. Hence, if we somehow want to access this information inside of a reaction, which your example does, then this would be the way to do it.

I'm not sure I agree that having this feature breaks modularity, however. You just retrieve a number. Not much you can do with it other than print it.

edwardalee:
Without this feature, there is no way to distinguish two reactors where the parent's parent's bank index differ. They will always have identical parameters.

lhstrh:
Yes, I agree. But the example you wrote breaks this principle. I think that if this feature is bad, then the example is also bad.

edwardalee:
If you rename `parents_bank_index` to `foo` and print `foo = %d`, would your objection to this example go away?

lhstrh:
That would just obfuscate what we're trying to do. If this is an "essential pattern," then let's be clear about it and allow it through this API call.

Either it's an antipattern and we should discourage it, or it's a pattern and we should encourage it.

lhstrh:
Closing this and will remove the feature.

PR #98, opened 2022-04-26
Refactoring, fixes, and tests for the graph package

mattchorlian:
This pull request adds unit tests and fixes several bugs.


Wonseo-C:
With @CloverCho 
1. There is some issue with a deep copy problem when we generate a new graph. We generate a new graph to check the connectivity, but this new graph affects an original (old) graph. So, we immediately fix this problem by removeEdgefunction. (Check the connectivity, then we remove the new Edge.)
2. We found that the causality graph was always empty in the direct feed-through condition. Because when we check the connectivity, the causality graph isn't created.
So, we fixed it by the solution which  checks the <code>dst</code> and <code>src</code> have the same reactor.
3. We modified the test case.

lhstrh:
For historical reasons, this PR is a bit of a hodgepodge, but with some careful review and all the tests passing, we should be able to merge this into `master`.

__tests__/CausalityGraph.test.ts
```diff
@@ -0,0 +1,89 @@
...
+        test('test if adding cyclic dependency is caught', () => {
+            expect(() => {
+                this._connect(this.reactor2.in, this.reactor1.out)
+            }).toThrowError("New connection introduces cycle.")
+        })
+    }
+}
+
+var app = new testApp()
+app._start()
```
lhstrh, line 89:
```suggestion
app._start()

```

__tests__/SimpleMutation.test.ts
```diff
@@ -0,0 +1,78 @@
...
+        this.start = new Starter(this);
+        this.reactor1 = new R1(this);
+        this.reactor2 = new R2(this);
+        this._connect(this.start.out, this.reactor1.in)
+        this._connect(this.reactor1.out, this.reactor2.in)
+    }
+}
+
+var app = new testApp()
+app._start()
```
lhstrh, line 78:
```suggestion
app._start()

```

__tests__/InvalidMutations.ts
```diff
@@ -0,0 +1,83 @@
...
+                        this.connect(__in1, __out2)
+                    }).toThrowError("New connection introduces direct feed through.")
+                })
+                test('expect error to be thrown on mutation creating loop', () => { 
+                    expect(() => {
+                        this.connect(__out2, __in1)
+                    }).toThrowError("New connection introduces direct feed through and cycle.")
+                })
+                let R2 = new R1(this.getReactor())
+                test('expect error on spawning and creating loop within a reactor', () => {
```
lhstrh, line 58: This test looks wrong. For the first connect statement to be legal, R1 has to be contained in the reactor that spawned it. For the second connect statement, R1 has to be upstream, which it can't be.

__tests__/SimpleMutation.test.ts
```diff
@@ -0,0 +1,78 @@
...
+    constructor(parent: Reactor|null) {
+        super(parent);
+        this.addMutation(
+            new Triggers(this.in),
+            new Args(this.in, this.out),
+            function(this, __in, __out) {
+                test('expect error to be thrown', () => { 
+                    expect(() => {
+                        this.connect(__out, __in)
+                    }).toThrowError("New connection introduces direct feed through.")
```
lhstrh, line 53: The error reported here should be that `__in` is already occupied.

src/core/graph.ts
```diff
@@ -0,0 +1,459 @@
...
+                    graph.delete(v);
+                }
+            }
+        } if (graph.size != 0) {
+            return false; // ERROR: cycle detected
+        } else {
+            return true;
+        }
+    }
+}
```
lhstrh, line 459:
```suggestion
}

```

lhstrh (commented):
Found two problems with the tests. Once resolved, we can go ahead and merge this.

__tests__/InvalidMutations.ts
lhstrh: Tagging @Wonseo-C

__tests__/SimpleMutation.test.ts
lhstrh: Tagging @Wonseo-C

Wonseo-C: https://github.com/lf-lang/reactor-ts/pull/98/commits/62b751a020f21a46b3eb8c5ef7d17f0912e9b7e3
I tried to fix the error message for this test case by adding the logic for the error in the ```canConnect``` function.

__tests__/InvalidMutations.ts
Wonseo-C: https://github.com/lf-lang/reactor-ts/pull/98/commits/58705b73556cde9e1538f951848f4a238262ab91
I made several modifications to the ```canConnect``` logic to fix the issues:
1) I updated the condition that checks for a 'destination port already occupied' error to include all cases, not just when ```this._runtime.isRunning() == false.```
2) I added a new condition to check for connections outside of the container.
3) I made some changes to the direct feed-through condition.


https://github.com/lf-lang/reactor-ts/pull/98/commits/638fd866a569b2da05ab78312979ab95c9650d12
I made an effort to fix the test case by removing the 'reaction2' and allowing for direct feed-through without the 'destination port is already occupied' condition. This was necessary because the previous version of the test case was causing issues with occupied destination ports.

lhstrh (approved):

Issue #101, opened 2022-05-11
Implementation of MSG_TYPE_STOP_REQUEST


hokeun:
Currently, federated TS does not support the [MSG_TYPE_STOP_REQUEST](https://github.com/lf-lang/reactor-c/blob/main/core/federated/net_common.h#L441) message.
We need to implement this so that we can stop the whole federation when a single federate requests stop.

hokeun:
@lhstrh , do you think you could add @ByeongGil-Jun to reactor-ts as a member? He will be working on this with me, and I plan to have him work on other reactor-ts issues as the next tasks. It would be much easier if he's a member and he can push to the origin branches. Thanks!

lhstrh:
@ByeongGil-Jun should have received an invitation.

ByeongGil-Jun:
@hokeun @lhstrh I showed and accepted an invitation. Thank you

ByeongGil-Jun:
@hokeun This is the line we talked about.
https://github.com/lf-lang/reactor-c/blob/e1e2bd3b16a3a450c266550ffe745170136fadf0/core/federated/net_common.h#L426

hokeun:
This will be addressed by https://github.com/lf-lang/reactor-ts/pull/111 and https://github.com/lf-lang/lingua-franca/pull/1302.

PR #111, opened 2022-07-19
Implement stop request in federated execution of TypeScript

ByeongGil-Jun:
Relevant issue: https://github.com/lf-lang/reactor-ts/issues/101

Relevant LF PR: https://github.com/lf-lang/lingua-franca/pull/1302

Add a logic for sending and propagating MSG_TYPE_STOP_REQUEST and MSG_TYPE_STOP_REQUEST_REPLY and receive MSG_TYPE_STOP_REQUEST_GRANTED
Discussion is needed to determine whether freeze federates or not.
Some function in reactor.ts is added and modified like getStartTag() and getStartTime() etc.


src/core/federation.ts
```diff
@@ -151,6 +151,65 @@ enum RTIMessageTypes {
      */
     MSG_TYPE_LOGICAL_TAG_COMPLETE = 9,
 
+    /////////// Messages used in lf_requestStop() ///////////////
```
lhstrh, line 154: This is not the right place to put this. How about we create a wiki page and link to it?

lhstrh (changes requested):
Looks good to me! My only request is that we start putting developer docs about federated execution on the wiki.

src/core/federation.ts
hokeun: Done win https://github.com/lf-lang/reactor-ts/pull/111/commits/b399e4fa9de86f341d1313af173450b79b225ce1!

@lhstrh, How do you think about we add @ByeongGil-Jun as a member of LF repo so that he can edit the wiki next time?

hokeun:
@ByeongGil-Jun, I granted you write access to the lingua-franca repo (https://github.com/lf-lang/lingua-franca/). From now on, you should be able to create branches and edit wiki pages here: https://github.com/lf-lang/lingua-franca/wiki

hokeun (approved):
I'm approving this to unblock @ByeongGil-Jun. @lhstrh, please feel free to comment on this PR if you come up with any suggestions later!

PR #114, opened 2022-08-03
Fix error in comparing intended tag and current tag in federated execution.

hokeun:
This error was discovered while testing the TypeScript chat application in https://github.com/lf-lang/examples-lingua-franca/pull/10/.

I also added unit tests so that we can avoid error in the future.

I found that the operators like <, <= do not work correctly in the generated .js code.

Raw error messages:
```
/Users/hokeunkim/Development/examples-lingua-franca/TypeScript/src-gen/ChatApplication/SimpleChat/dist/core/action.js:112
            throw new Error("Intended tag must be greater than current tag. Intended tag" + intendedTag + " Current tag: " + this.action.runtime.util.getCurrentTag());
            ^

Error: Intended tag must be greater than current tag. Intended tag((1659547361 secs; 115322112 nsecs), 1) Current tag: ((1659547361 secs; 45592064 nsecs), 0)
    at Object.schedule (/Users/hokeunkim/Development/examples-lingua-franca/TypeScript/src-gen/ChatApplication/SimpleChat/dist/core/action.js:112:19)
    at RTIClient.<anonymous> (/Users/hokeunkim/Development/examples-lingua-franca/TypeScript/src-gen/ChatApplication/SimpleChat/dist/core/federation.js:1319:68)
    at RTIClient.emit (node:events:526:28)
    at RTIClient.handleSocketData (/Users/hokeunkim/Development/examples-lingua-franca/TypeScript/src-gen/ChatApplication/SimpleChat/dist/core/federation.js:698:22)
    at Socket.emit (node:events:526:28)
    at addChunk (node:internal/streams/readable:315:12)
    at readableAddChunk (node:internal/streams/readable:289:9)
    at Socket.Readable.push (node:internal/streams/readable:228:10)
    at TCP.onStreamRead (node:internal/stream_base_commons:190:23)
```

This also increases test coverage for action.ts (68.63% -> 86.27%):

Before:
Screen Shot 2022-08-03 at 10 46 57 AM

After:
Screen Shot 2022-08-03 at 10 46 32 AM

### DIFF
--- a/__tests__/action.test.ts
+++ b/__tests__/action.test.ts
@@ -1,0 +1,113 @@
+import {Action, FederatePortAction, App, Origin, TimeValue, Triggers, Args, Tag} from '../src/core/internal';
+
+let intendedTagDelay: TimeValue | undefined
+let intendedTagMicrostepDelay: number
+let startUpTag: Tag
+
+class ReactorWithFederatePortAction extends App {
+    a = new Action<number>(this, Origin.logical, TimeValue.nsec(10))
+    f = new FederatePortAction<number>(this, Origin.logical)
+    
+    constructor () {
+        super(TimeValue.msec(1));
+        this.addReaction(
+            new Triggers(this.startup),
+            new Args(this.schedulable(this.a)),
+            function(this, a){
+                startUpTag = this.util.getCurrentTag()
+                a.schedule(0, 0);
+            }
+        );
+
+        this.addReaction(
+            new Triggers(this.a),
+            new Args(this.schedulable(this.f)),
+            function(this, f){
+                let intendedTag: Tag | undefined
+                if (intendedTagDelay === undefined) {
+                    intendedTag = undefined
+                } else {
+                    intendedTag = startUpTag.getLaterTag(intendedTagDelay).getMicroStepsLater(intendedTagMicrostepDelay)
+                }
+                f.schedule(0, 0, intendedTag);
+            }
+        );
+    }
+
+    public setLastTagProvisional(value: boolean) {
+        this._isLastTAGProvisional = value
+    }
+}
+
+describe('Intended tag tests', function () {
+    it('Undefined inteded tag', function() {
+        intendedTagDelay = undefined
+        let app = new ReactorWithFederatePortAction()
+        expect(() => app._start()).toThrowError("FederatedPortAction must have an intended tag from RTI.")
+    })
+    
+    it('Intended tag smaller than current tag', function() {
+        intendedTagDelay = TimeValue.nsec(9)
+        intendedTagMicrostepDelay = 0
+        let app = new ReactorWithFederatePortAction()
+        app.setLastTagProvisional(false)
+        expect(() => app._start()).toThrowError("Intended tag must be greater than current tag. Intended tag")
+    })
+    
+    it('Intended tag equal to current tag', function() {
+        intendedTagDelay = TimeValue.nsec(10)
+        intendedTagMicrostepDelay = 1
+        let app = new ReactorWithFederatePortAction()
+        app.setLastTagProvisional(false)
+        expect(() => app._start()).toThrowError("Intended tag must be greater than current tag. Intended tag")
+    })
+    
+    it('Intended tag greater than current tag', function() {
+        intendedTagDelay = TimeValue.nsec(11)
+        intendedTagMicrostepDelay = 0
+        let app = new ReactorWithFederatePortAction()
+        app.setLastTagProvisional(false)
+        app._start()
+    })
+
+    it('Intended tag greater than current tag by microstep', function() {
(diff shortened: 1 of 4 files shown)


__tests__/action.test.ts
```diff
@@ -0,0 +1,113 @@
...
+        let app = new ReactorWithFederatePortAction()
+        expect(() => app._start()).toThrowError("FederatedPortAction must have an intended tag from RTI.")
+    })
+    
+    it('Intended tag smaller than current tag', function() {
+        intendedTagDelay = TimeValue.nsec(9)
+        intendedTagMicrostepDelay = 0
+        let app = new ReactorWithFederatePortAction()
+        app.setLastTagProvisional(false)
+        expect(() => app._start()).toThrowError("Intended tag must be greater than current tag. Intended tag")
```
lhstrh, line 54: The thrown string appears to be incomplete.

__tests__/action.test.ts
```diff
@@ -0,0 +1,113 @@
...
+    })
+    
+    it('Intended tag equal to current tag', function() {
+        intendedTagDelay = TimeValue.nsec(10)
+        intendedTagMicrostepDelay = 1
+        let app = new ReactorWithFederatePortAction()
+        app.setLastTagProvisional(false)
+        expect(() => app._start()).toThrowError("Intended tag must be greater than current tag. Intended tag: ")
```
lhstrh, line 62: The thrown string appears to be incomplete.

__tests__/action.test.ts
```diff
@@ -0,0 +1,113 @@
...
+    })
+    
+    it('Intended tag greater than current tag', function() {
+        intendedTagDelay = TimeValue.nsec(11)
+        intendedTagMicrostepDelay = 0
+        let app = new ReactorWithFederatePortAction()
+        app.setLastTagProvisional(false)
+        app._start()
```
lhstrh, line 70: This doesn't appear to test anything...

__tests__/action.test.ts
```diff
@@ -0,0 +1,113 @@
...
+    })
+
+    it('Intended tag greater than current tag by microstep', function() {
```
lhstrh, line 73: This doesn't appear to test anything...

__tests__/action.test.ts
```diff
@@ -0,0 +1,113 @@
...
+    })
+})
+
+describe('Intended tag tests when TAG is provisional', function () {
+    it('Intended tag smaller than current tag', function() {
+        intendedTagDelay = TimeValue.nsec(9)
+        intendedTagMicrostepDelay = 0
+        let app = new ReactorWithFederatePortAction()
+        app.setLastTagProvisional(true)
+        expect(() => app._start()).toThrowError("Intended tag must be greater than or equal to current tag, when the last TAG is provisional. Intended tag: ")
```
lhstrh, line 88: The thrown string appears to be incomplete.

__tests__/action.test.ts
```diff
@@ -0,0 +1,113 @@
...
+    })
+    
+    it('Intended tag equal to current tag', function() {
```
lhstrh, line 91: The next three tests also have no "expect". What do they test?

lhstrh (commented):
This looks very good! It's surprising these problems didn't surface earlier... I left a few comments.

__tests__/action.test.ts
hokeun: This is meant to match the prefix because the actual time value can change (because we cannot set the current Tag outside of App). Would you prefer the regular expressions?

__tests__/action.test.ts
hokeun: Ah, these tests are to make sure there's no exception. Could we explicitly expect no exception?

__tests__/action.test.ts
lhstrh: I wasn't aware that matching a prefix suffices to pass the test. Is that what `toThrowError` does?

hokeun: It looks like `toThrowError` performs substring matching.

https://jestjs.io/docs/expect

**Screenshot below**:
Screen Shot 2022-08-04 at 9 05 45 PM

-----------
I think this should be enough as the error message is quite specific, however, I'm open to suggestions. :)

__tests__/action.test.ts
lhstrh: I think that would be better. That way it is more explicit what is expected. Adding `.not.toThrow()` should do the trick.

hokeun: Done. Thanks for the suggestion!

__tests__/action.test.ts
hokeun: Thanks for catching it! I believe all comments are addressed now.

lhstrh (approved):
Awesome! Thanks for the fixes!

PR #115, opened 2022-08-04
Fix errors with reaction orders.

hokeun:
Fix errors with reaction orders by making sure that all events at the certain tag are collected before executing reactions. (Previously, reactions were executed even when there are other events at the same tag were not collected.)

This was discovered while working on port absent messages with @ByeongGil-Jun.

Here is how to replicate the error:

C target LF program:
https://github.com/lf-lang/examples-lingua-franca/blob/port-absent/C/src/PortAbsentTest.lf

TS target LF program:
https://github.com/lf-lang/examples-lingua-franca/blob/port-absent/TypeScript/src/PortAbsentTest.lf

If you run both of the programs above, you will notice that the C version executes reactions in the order they're defined. However, the TS version executes reactions in a different order.

### DIFF
--- a/src/core/reactor.ts
+++ b/src/core/reactor.ts
@@ -2062,9 +2062,6 @@ export class App extends Reactor {
                     // Look at the next event on the queue.
                     nextEvent = this._eventQ.peek();
                 }
-
-                // React to all the events loaded onto the reaction queue.
-                this._react()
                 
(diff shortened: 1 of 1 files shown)


lhstrh (approved):
This looks right to me! Thanks for this fix!!

Issue #105, opened 2022-07-05
Implementation of MSG_TYPE_PORT_ABSENT in reactor-ts


hokeun:
MSG_TYPE_PORT_ABSENT is a message sent by an output control reaction when the output is going to be absent.

See [here](https://github.com/lf-lang/reactor-c/blob/main/core/federated/net_common.h#L609) for more details.

hokeun:
This is a comment where port absent message was mentioned: https://github.com/lf-lang/lingua-franca/pull/538#discussion_r715972519

hokeun:
FYI @ByeongGil-Jun on federated execution AST transformation: https://github.com/lf-lang/lingua-franca/discussions/1212

hokeun:
FYI @ByeongGil-Jun on network control reactions: https://github.com/lf-lang/lingua-franca/discussions/608

hokeun:
@lhstrh This is the tracking issue for port absent messages. Thanks!

ByeongGil-Jun:
Below is the logic to execute reactions considering network control reactions and port absent messages in TypeScript federated execution. The logic introduces two separate queues for reactions `waitingReactionQ` and `readyReactionQ`, (previously just one `reactionQ`). 

`LastKnownStatusTag`: Tag indicating the last time when the port received port absent message or tagged message.
`waitingReactionQ`: Queue of reactions that are not ready to run. We should pop the reaction from this queue and push it into the `readyReactionQ`.
`readyReactionQ`: Queue of the reactions ready to run. If the reaction is the network input control reaction and the return value is waiting, we should insert this reaction again in `waitingReactionQ` as well as the reactions to run after the reaction that must wait.
First, insert all reactions that have a network input port into `waitingReactionQ`, (no network control reaction added yet at this point).

This is a while loop. This loop exits only when `waitingReactioinQ` and `reactionReactionQ` are empty.
Check all reactions in the `waitingReactionQ` in order and move them to the `readyReactionQ`. If any port has unknown status, the reaction can’t be inserted into the `readyReactionQ`. In this case, we should insert a `networkInputControlReaction` into the `readyReactionQ`.
[...]

PR #119, opened 2022-08-24
Implementation of `portAbsent` and `outputControlReaction` in federated execution of TypeScript

byeong-gil:
Relevant issue: #105

Relevant LF PR: [lf-lang/lingua-franca#1333](https://github.com/lf-lang/lingua-franca/pull/1333)

Add logic for sending and handling MSG_TYPE_PORT_ABSENT and triggering output control reactions.

Some functions of functions are added and modified like sendRTIPortAbsent(), registerOutputControlReaction() etc.

The logic for handling input control reaction should be added and this will be updated soon.


src/core/federation.ts
```diff
@@ -828,6 +864,23 @@ class RTIClient extends EventEmitter {
                     bufferIndex += 13;
                     break;
                 }
+                case RTIMessageTypes.MSG_TYPE_PORT_ABSENT: {
+                    // The next 2 bytes is the port id.
```
hokeun, line 868: NIT: next 2 bytes is -> next 2 bytes are

src/core/federation.ts
```diff
@@ -1178,6 +1254,25 @@ export class FederatedApp extends App {
         this.rtiClient.sendRTIStopRequestReply(tag);
     }
 
+    /**
+     * Send a port absent message to federate with fed_ID, informing the
```
hokeun, line 1258:
```suggestion
     * Send a port absent message to the destination port, informing the
```

src/core/federation.ts
```diff
@@ -1178,6 +1254,25 @@ export class FederatedApp extends App {
...
+     * remote federate that the current federate will not produce an event
+     * on this network port at the current logical time.
+     * 
+     * @param additional_delay The offset applied to the timestamp
```
hokeun, line 1262:
```suggestion
     * @param additionalDelay The offset applied to the timestamp
```

src/core/federation.ts
```diff
@@ -1178,6 +1254,25 @@ export class FederatedApp extends App {
...
+     *  using after. The additional delay will be greater or equal to zero
+     *  if an after is used on the connection. If no after is given in the
+     *  program, NEVER is passed.
```
hokeun, line 1265: `If no after is given in the program, NEVER is passed` -> This statement doesn't seem true from the implementation. Maybe we should update the implementation to match this description?

byeong-gil:
This PR works well except for tests that failed in the `fed-gen` branch. 
We can see reactors send and receive `port-absent` messages every time when the port doesn't make outputs.

src/core/reactor.ts
```diff
@@ -1819,6 +1823,22 @@ export class App extends Reactor {
...
+     * Send a port absent message to federate with fed_ID, informing the
+     * remote federate that the current federate will not produce an event
+     * on this network port at the current logical time.
+     * 
+     * @param additional_delay The offset applied to the timestamp
+     *  using after. The additional delay will be greater or equal to zero
+     *  if an after is used on the connection. If no after is given in the
+     *  program, NEVER is passed.
+     * @param destFederatedID The fed ID of the receiving federate.
+     * @param destPortID The ID of the receiving port.
```
hokeun, line 1836: I think we should update these comments as this function is just to be overridden by the subclass.

hokeun (approved):
Looks good to me. Just left minor suggestions.

byeong-gil:
> Looks good to me. Just left minor suggestions.

These suggestions are all fixed in this [commit](https://github.com/lf-lang/reactor-ts/commit/20d16b96d5848ec9a721b1f19144870f7836cee8).

src/core/reactor.ts
hokeun: I fixed this myself.

PR #122, opened 2022-08-28
Reference reactor-ts as module

petervdonovan:
Companion PR: https://github.com/lf-lang/lingua-franca/pull/1322


src/core/port.ts
```diff
@@ -34,7 +34,7 @@ export abstract class Port<T extends Present> extends Trigger {
     public isPresent() {
 
         Log.debug(this, () => "In isPresent()...")
-        Log.debug(this, () => "value: " + this.value);
+        Log.debug(this, () => "value: " + this.value?.toString);
```
lhstrh, line 37: Shouldn't this be `toString()` (parentheses missing).

package.json
```diff
@@ -84,7 +83,10 @@
         "url": "http://people.eecs.berkeley.edu/~marten"
     },
     "contributors": [
+        "Yunsang Cho",
+        "Wonseo Choi",
```
lhstrh, line 87:
```suggestion
        "Wonseo Choi",
        "Peter Donovan",
```

lhstrh (approved):
Looks good to me!

package.json
```diff
@@ -1,52 +1,51 @@
...
-        "typescript": "^4.3.5"
+        "typedoc": "0.23.11"
     },
-    "files": [
-        "lib"
-    ],
     "scripts": {
         "check-types": "tsc",
-        "build": "rm -rf dist && babel src --out-dir dist --extensions .ts",
+        "install": "npx rimraf lib && npx babel src --out-dir lib --extensions .ts",
```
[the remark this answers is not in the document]
petervdonovan: I'm confused. Previously, we were using `rm`, which was not portable; this change **introduces** the use of `rimraf`.

src/core/port.ts
```diff
@@ -34,7 +34,7 @@ export abstract class Port<T extends Present> extends Trigger {
     public isPresent() {
 
         Log.debug(this, () => "In isPresent()...")
-        Log.debug(this, () => "value: " + this.value);
+        Log.debug(this, () => "value: " + this.value?.toString());
```
suyourice, line 37: Why isn't this part like `value.toString()` ?

package.json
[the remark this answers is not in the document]
lhstrh: Sorry -- misread the diff!

src/core/port.ts
petervdonovan: [The question mark handles the fact that `this.value` is nullable](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining)

suyourice: Ah thank you. :)

PR #124, opened 2022-08-31
Address CI failure in documentation generation.

petervdonovan:
I do not know why TypeDoc couldn't find our entry point inside the
src folder, but I guess it isn't suprising since I did do a version
bump. I also had to skip lib checking due to issues with Mocha:
https://stackoverflow.com/questions/55680391/typescript-error-ts2403-subsequent-variable-declarations-must-have-the-same-typ

(Note that I think it's a good idea to skip lib checking regardless, since the code we depend on in `reactor-ts` is maintained by other people and should not be our problem. It will save time not to have `tsc` check all that.)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -31,7 +31,7 @@
                 "microtime": "^3.0.0",
                 "nanotimer": "^0.3.15",
                 "rimraf": "^3.0.2",
-                "typescript": "~4.4.4",
+                "typescript": "~4.8.2",
                 "ulog": "^1.1.0",
                 "uuid": "^8.3.2"
             },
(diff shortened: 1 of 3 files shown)


petervdonovan:
The amount of time being taken by CI here is outlandish. In Ubuntu tests finished in 1 hour, but it has taken almost 3 hours in macOS and over 4 hours on Windows, and it is not even finished yet. I will be interested to see if this is reproducible -- it seems like `npm` depends heavily on the network.

lhstrh:
I agree. CI is getting too slow to be workable...

lhstrh:
On top of that, we get a lot of intermittent failures due to network issue. Right now, for instance, the Maven build is failing because Kiel's update site is unreachable.

lhstrh (approved):

petervdonovan:
I'm concerned that this change made CI slower on Mac (or maybe on all platforms). I haven't done a statistical analysis (anyway, a statistical analysis would be hard due to confounders) but a quick glance at the history suggests that the TS tests went from perhaps 2 hours to perhaps 4 hours in the past few days. I hesitate to spend time fixing this -- I just wanted to acknowledge for the record that I might have exacerbated this existing problem.

PR #127, opened 2022-10-22
Cleanup

lhstrh:
- [x] Remove superfluous `.d.ts` files and get types from npm instead
 - [x] Remove deprecated code and associated tests
 - [x] Attempt to generate `lib/core/index.d.ts` so that types are included in the npm mpdule (also see https://cameronnokes.com/blog/the-30-second-guide-to-publishing-a-typescript-package-to-npm/)


petervdonovan:
The issue that is causing the TypeScript tests to fail right now looks like it could be the same issue that I referenced previously in [this comment](https://github.com/lf-lang/lingua-franca/pull/1322#issuecomment-1223183810).

lhstrh:
> The issue that is causing the TypeScript tests to fail right now looks like it could be the same issue that I referenced previously in [this comment](https://github.com/lf-lang/lingua-franca/pull/1322#issuecomment-1223183810).

Yes, it sure looks like it. I think this is a shortcoming, not an intended feature of TS. Or is it?

petervdonovan:
Yes, based on the [linked issue](https://github.com/microsoft/TypeScript/issues/35822) it looks like we are not the only ones who experience this as a shortcoming.

lhstrh:
Looks like this passes. OK to merge, @petervdonovan?

petervdonovan (approved):

PR #128, opened 2022-10-26
Reduction in time spent on Lingua Franca tests

lhstrh:
This PR simplifies the GitHub Actions configuration and only runs Lingua Franca tests on Linux.


petervdonovan (approved):
This looks right to me!

PR #135, opened 2023-02-17
Add the microstep only when the actions are scheduled at the current tag

byeong-gil:
For example, when action A is scheduled by `actions.a.schedule(TimeValue.sec(1), null)` at the tag (1 sec, 0), it is scheduled at the tag (2 sec, 1).
[reactor-ts/src/core/action.ts/#L97](https://github.com/lf-lang/reactor-ts/blob/0e9fba3dad5d45c304e4b92b233eb8faa549688a/src/core/action.ts#L97) is the reason of that. 

We have to increment the microstep only when we try to schedule an action at the current tag.


lhstrh (approved):
:+1:

lhstrh:
Looks good, but I'm surprised to see the unit tests failing...

byeong-gil:
> Looks good, but I'm surprised to see the unit tests failing...

Yeah right... I have to debug.

lhstrh:
Looks like the test might be wrong? At least we have better coverage than I thought :-)

byeong-gil:
Yeah it's good for reactor-ts :). I think that else if statement cannot prevent the schedule at the exact current tag. I'm not sure though...

lhstrh (approved):

PR #136, opened 2023-02-23
Implementation of the disconnect function

Wonseo-C:
I have implemented the ```_disconnect()``` function in the Reactor class, which can now be utilized in Mutation. Additionally, I have added a new test called ```disconnect.test.ts``` to confirm that connections are successfully disconnected. Furthermore, I have included the disconnect portion in the ```simple.ts``` test.


__tests__/disconnect.test.ts
```diff
@@ -0,0 +1,67 @@
...
+            new Triggers(this.in1),
+            new Args(this.in1, this.in2, this.out2),
+            function(this, __in1, __in2, __out2) {
+                let R2 = new R1(this.getReactor())
+                test('expect that disconnecting an existing connection will not result in an error being thrown', () => {
+                    expect(() => {
+                        this.connect(R2.out2, R2.in2)
+                        this.disconnect(R2.out2, R2.in2)
+                        this.connect(R2.out2, R2.in2)
+                    }).not.toThrow("Destination port is already occupied.")
```
lhstrh, line 46: Shouldn't it be expected that no exception is thrown at all?

src/core/reactor.ts
```diff
@@ -1279,11 +1287,20 @@ protected _getFirstReactionOrMutation(): Reaction<any> | undefined {
      * @param src
      * @param dst
      */
-    private _disconnect(src: Port<Present>, dst: Port<Present>) {
-        Log.debug(this, () => "disconnecting " + src + " and " + dst);
-        //src.getManager(this.getKey(src)).delReceiver(dst);
-
-
+    protected _disconnect(src: IOPort<Present>, dst: IOPort<Present>) {
+        if (src === undefined || src === null) {
```
lhstrh, line 1291: I don't think it is possible to pass in `undefined` so this check shouldn't be necessary?

src/core/reactor.ts
```diff
@@ -1279,11 +1287,20 @@ protected _getFirstReactionOrMutation(): Reaction<any> | undefined {
...
+            throw new Error("Cannot disconnect unspecified source");
+        }
+        if (dst === undefined || dst === null) {
```
lhstrh, line 1294: Idem

src/core/reactor.ts
```diff
@@ -1293,6 +1310,31 @@ protected _getFirstReactionOrMutation(): Reaction<any> | undefined {
...
+        src.getManager(this._getKey(src)).delReceiver
+            (writer as WritablePort<S>);
+        this._dependencyGraph.removeNode(src);
+        this._dependencyGraph.removeNode(dst);
+        
+    }
+
+    public canDisconnect<R extends Present, S extends R>
+        (src: IOPort<S>, dst: IOPort<R>) {
+        if (src === dst) {
```
lhstrh, line 1327: I don't think we need to handle this case separately.

lhstrh (commented):
Looks good! I left some minor comments...

src/core/reactor.ts
Wonseo-C: I redefined the ```src``` and ```dst``` types to match those used in the '_connect' function.

src/core/reactor.ts
Wonseo-C: I merged the condition for ```canDisconnect``` into the ```_disconnect``` function to simplify it and allow for ```_uncheckedDisconnect``` to be executed.

__tests__/disconnect.test.ts
Wonseo-C: Yes, that sounds reasonable. I removed the message that was in the ```toThrow``` statement.

lhstrh (approved):

src/core/reactor.ts
```diff
@@ -1309,24 +1306,27 @@ protected _getFirstReactionOrMutation(): Reaction<any> | undefined {
...
         } else {
-            let nodes = this._dependencyGraph.getEdges(src);
-            for (let node of nodes) {
-                if (node instanceof IOPort) {
-                    let writer = src.asWritable(this._getKey(node));
-                    src.getManager(this._getKey(src)).delReceiver
-                        (writer as WritablePort<S>);
+            let nodes = this._dependencyGraph.getBackEdges(src);
+            if (nodes.size == 0) {
+                throw Error("There are no destinations to disconnect")
```
lhstrh, line 1319: I don't think this is an error. There's just nothing to do, so we can return. Let's not error.

src/core/reactor.ts
```diff
@@ -398,6 +398,14 @@ export abstract class Reactor extends Component {
             }
         }
 
+        public disconnect(src: IOPort<Present>, dst?: IOPort<Present>): void {
+            if (src instanceof IOPort && (dst === undefined || dst instanceof IOPort)) {
+                return this.reactor._disconnect(src, dst);
+            } else {
+                // ERROR
```
hokeun, line 405:
```suggestion
                // FIXME: Add an error reporting mechanism such as an exception.
```

src/core/reactor.ts
```diff
@@ -1287,22 +1292,42 @@ protected _getFirstReactionOrMutation(): Reaction<any> | undefined {
 
 
     /**
-     *
+     * Delete the connection between the source and destination nodes. 
+     * If the destination node is not specified, all connections from the source node to any other node are deleted.
      * @param src
      * @param dst
```
hokeun, line 1298:
```suggestion
     * @param src Source port of connection to be disconnected.
     * @param dst Destination port of connection to be disconnected. If undefined, disconnect all connections from the source port.
```

hokeun (approved):
Great work!

src/core/reactor.ts
Wonseo-C: Yes, I agree! I fixed the issue.

PR #125, opened 2022-09-01
Address`after` delay in Federated Execution

byeong-gil:
Relevant LF PR: 

This PR adds logic for handling `after` delay in federated execution.
`additionalDelay` is reflected when sending `taggedMessage`.

Some functions in time.ts are added and modified like getLaterTag() and isNEVER() etc.

PR #138, opened 2023-02-27
Send appropriate upstream connection delays to the RTI

byeong-gil:
The previous version of this PR is [reactor-ts/pull/125](https://github.com/lf-lang/reactor-ts/pull/125).

This PR addresses This [FIXME](https://github.com/lf-lang/reactor-ts/blob/be7b4f18ce8e40098867a25fe4f59929862038a8/src/core/federation.ts#L1153). 

This PR adds an element `upstreamConnectionDelays` to the `FederateConfig`.
Also, it determines the minimum delays from each upstream federate and adds it to the list, [`upstreamFedDelays`](https://github.com/lf-lang/reactor-ts/blob/be7b4f18ce8e40098867a25fe4f59929862038a8/src/core/federation.ts#L987).

### DIFF
--- a/lingua-franca-ref.txt
+++ b/lingua-franca-ref.txt
@@ -1,1 +1,1 @@
-master
+ts-upstream-delays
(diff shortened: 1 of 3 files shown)


lhstrh (approved):
:+1:

PR #141, opened 2023-02-28
Simplification of build configuration

lhstrh:
- removal of unused dependencies
- removal of dependency on `babel` (build using `tsc`)
- use of `ts-jest` instead of `babel` for testing


axmmisaka:
This is failing because it is using npm package `@lf-lang/reactor-ts` that relies on `node_modules/@babel/*` which cannot compile without babel - this feels a bit circular (`@lf-lang/reactor-ts`'s dependency will be changed as this PR is merged); let me see what we should do

lhstrh:
Jest doesn't *need* Babel but we used to have it configured to use Babel. There are alternative configurations, I believe, using `ts-jest`, but I didn't have the time to investigate this further...

axmmisaka:
Some of the stuff we use does seem to rely on babel related packages: https://www.npmjs.com/package/jest-snapshot?activeTab=dependencies

axmmisaka:
@lhstrh Can you approve #141 's CICD run later, see if it works......

petervdonovan:
> src/c.ts(17,7): error TS2741: Property 'upstreamConnectionDelays' is missing in type '{ dependsOn: never[]; executionTimeout: undefined; fast: false; federateID: number; federationID: string; keepAlive: false; minOutputDelay: undefined; networkMessageActions: string[]; rtiHost: string; rtiPort: number; sendsTo: number[]; }' but required in type 'FederateConfig'.
    ERROR: Compiler failed with the following errors:

 >   src/p.ts(17,7): error TS2741: Property 'upstreamConnectionDelays' is missing in type '{ dependsOn: number[]; executionTimeout: undefined; fast: false; federateID: number; federationID: string; keepAlive: false; minOutputDelay: undefined; networkMessageActions: string[]; rtiHost: string; rtiPort: number; sendsTo: never[]; }' but required in type 'FederateConfig'.

It looks like something did not type check. It is surprising to me that this type error is surfacing only now, but it does look like it is more than just a flaky test.

lhstrh:
This looks like something @byeong-gil would know how to fix. On the upside, it looks like test time is down at least somewhat (from 45 min to ~30).

axmmisaka:
Oh lord, I just realised I was staring at the old CICD output......
Another reason why we should switch from `babel` to `tsc`; 
BTW, what's the significance of 24e00d9f? We appear to be using ES6 pattern (if I'm not mistaken) everywhere

petervdonovan:
> We appear to be using ES6 pattern (if I'm not mistaken) everywhere

This is always a source of confusion. TS is a superset of JS, but it is like a different language in the sense that in order to run as JS it has to be totally transpiled. So even though the syntax in TS files might look like ES modules syntax, the format of the generated JS files can be whatever we like. It is preferable to transpile to CommonJS because CommonJS is older than ES modules and is therefore compatible with more things.

axmmisaka:
I see, thanks! I understand that is transpiler settings but I somehow had the impression that we write ts in the way we want to transpile......

.github/workflows/ci.yml
```diff
@@ -41,6 +41,8 @@ jobs:
           fetch-depth: 0
       - name: Prepare build environment
         uses: ./lingua-franca/.github/actions/prepare-build-env
+        with:
```
lhstrh, line 44:
```suggestion
```

.github/workflows/ci.yml
```diff
@@ -41,6 +41,8 @@ jobs:
...
+          lf-dir: "lingua-franca"
```
lhstrh, line 45:
```suggestion
```

lhstrh:
Looks like we're back to only the Docker tests failing. Could you have a look @byeong-gil?

lhstrh:
It seems that the Docker tests simply use the latest release of `reactor-ts`. I'm not sure how to fix this.

byeong-gil:
> It seems that the Docker tests simply use the latest release of `reactor-ts`. I'm not sure how to fix this.

Maybe this [PR#1633](https://github.com/lf-lang/lingua-franca/pull/1633) can fix this?

axmmisaka:
Trying to revive this - I am trying to see why was #1633 closed?
Immature and uneducated question:
My impression is that we might need to change CI config for this repo as well - modify `package.json` to use the branch in the PR to run integrated test (https://github.com/lf-lang/lingua-franca/blob/2b75fe73ffb2f395db6f391d991adf3fc643612f/org.lflang/src/org/lflang/generator/ts/TSGenerator.kt#L190-L192)? I'm not sure what the default behaviour of `checkout` is: https://github.com/actions/checkout/issues/1108; https://github.com/lf-lang/reactor-ts/blob/5a5f51620f9d6fe0c01076e75ec8e39d6cefc50c/.github/workflows/ci.yml#L28-L31
Suppose it works, however, my confusion is that the runtime is correctly set to the current branch (https://github.com/lf-lang/reactor-ts/actions/runs/4382513810/jobs/7671643961#step:7:206), but it appears to me (which might be totally wrong) that we are fetching a published version in the failed test (https://github.com/lf-lang/reactor-ts/actions/runs/4382513810/jobs/7671643961#step:7:318).

byeong-gil:
>my confusion is that the runtime is correctly set to the current branch (https://github.com/lf-lang/reactor-ts/actions/runs/4382513810/jobs/7671643961#step:7:206), but it appears to me (which might be totally wrong) that we are fetching a published version in the failed test (https://github.com/lf-lang/reactor-ts/actions/runs/4382513810/jobs/7671643961#step:7:318).

I think that's because ``` dependencies:
    + @lf-lang/reactor-ts 0.3.2``` in the failed test (https://github.com/lf-lang/reactor-ts/actions/runs/4382513810/jobs/7671643961#step:7:318) is from https://github.com/lf-lang/reactor-ts/blob/2dae069b1421227dc6b6f909a74d8e80dc111b2e/package.json#L3

The above code is a part of this [`package.json`](https://github.com/lf-lang/reactor-ts/blob/cleanup-deps/package.json#L3) file. This `package.json` file is in the `cleanup-deps` branch but it says its version 0.3.2.


Additionally, I think this branch can pass the test after aligning this `reactor-ts/cleanup-deps` branch and [lingua-franca/faster-ts](https://github.com/lf-lang/lingua-franca/tree/faster-ts) branch with the master branches. However, there is a [conflict](https://github.com/lf-lang/lingua-franca/pull/1611/conflicts) on the branch `faster-ts`. Could anyone resolve that conflict and merge the master branch to the `faster-ts`?

lhstrh (commented):
Looks like this is good to go now! @petervdonovan?

petervdonovan (approved):
LGTM

PR #142, opened 2023-03-05
Remove Babel dependencies

axmmisaka:
Hopefully fixes https://github.com/lf-lang/reactor-ts/pull/141
Stupid as it seems, the reason appears to be that we are introducing dependencies that are useless

But we probably can't eliminate babel in dev environment......

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,12 +1,2 @@
 {
-  "presets": [
-    "@babel/preset-typescript"
-    //,"@babel/preset-env"
-  ],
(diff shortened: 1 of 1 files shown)


lhstrh:
Oh, fantastic! :thinking: Could we just remove `.babelrc` altogether?

lhstrh:
Merging this because it is progress. Let me also add you to the repo so you can push to the `cleanup-deps` branch directly.

Issue #86, opened 2022-03-01
Redesign how command-line-args and command-line-usage modules work


hokeun:
For example, cli.ts has

```
import commandLineArgs from 'command-line-args';
import commandLineUsage from 'command-line-usage';
```

which are not directly used inside cli.ts, but maybe in the generated code.
We could design this better.

@lhstrh Please feel free to edit the description here. :)

lhstrh:
I agree. These import statements should be included in the generated code instead. That will also remove a dependency from this project...

@axmmisaka are you interested in picking this up? Should be an easy fix...

axmmisaka:
What I tried to do is use this (https://github.com/lf-lang/vscode-lingua-franca/blob/main/tslint.json), convert it to ESLint config and run ESLint on the whole repo. After the exam tomorrow I think it might be a good idea to fix all of them and fix things broken by the linter

lhstrh:
@axmmisaka, there is a pending PR #83 that uses Prettier instead of ESLint.

lhstrh:
Also, this issue isn't really about formatting or linting but about removing a dependency from `reactor-ts` that should be introduced by the code generator.

lhstrh:
> > @axmmisaka, there is a pending PR #83 that uses Prettier instead of ESLint.
> 
> If I recall correctly, ESLint and Prettier do different things
> 
True. I think it's a good idea to also set up ESLint and fix those problems, but let's discuss that in a separate issue/PR.

axmmisaka:
> > > @axmmisaka, there is a pending PR #83 that uses Prettier instead of ESLint.
> > 
> > 
> > If I recall correctly, ESLint and Prettier do different things
> 
> True. I think it's a good idea to also set up ESLint and fix those problems, but let's discuss that in a separate issue/PR.

I'm a bit curious, though, what this issue is trying to address - it doesn't appear to be a design problem if it's not used anywhere (in contrast, it could potentially be a design problem if, say, it's used somewhere, but solely as a placeholder)

PR #83, opened 2022-02-19
Support for auto-formatting using Prettier

lhstrh:
To run: `npm run format`

PR #146, opened 2023-03-13
Add eslintrc and lint some files

axmmisaka:
I'm not 100% sure how the "design" (https://github.com/lf-lang/reactor-ts/issues/86) here is supposed to work, but I tried to run eslint and it looks really bad (lots of `unused-var` and `eqeqeq` issue); most of them are, sadly, not auto-fixable.

I fixed some, but not sure what route do we plan to proceed... Posting a draft PR here first

Speaking of which, we should also run `prettier` or similar things


lhstrh:
We have #83 but have been reluctant to merge it because of outstanding PRs that would end up with conflicts...

lhstrh:
The reason why CI fails is that it expects `lint-wish-eslint` to be a branch on `lf-lang/reactor-ts`. We can either try to fix this, or you can create a branch in this repo instead. If you don't have permission to do this I can grant it to you.

lhstrh:
This patch might do the trick: https://github.com/lf-lang/reactor-ts/commit/91fd3b38dadb9293a77fc9793909d45ffe8aafad

axmmisaka:
Closing this, when #83 is merged open a new one based off of that.

PR #149, opened 2023-04-16
Unnecessary field `rtiSynchronized` removed

byeong-gil:
The variable "rtiSynchronized" is only used at [https://github.com/lf-lang/reactor-ts/blob/b5a29738c38222a44b64fc36091e3679e15fe1f7/src/core/federation.ts#L1059](https://github.com/lf-lang/reactor-ts/blob/b5a29738c38222a44b64fc36091e3679e15fe1f7/src/core/federation.ts#L1059)

This prevents the upstream federate with a physical action to schedule dummy events to advance the downstream federate's tag.

I think we can simply remove this concept. The code above is a part of [`_canProceed()`](https://github.com/lf-lang/reactor-ts/blob/b5a29738c38222a44b64fc36091e3679e15fe1f7/src/core/federation.ts#L1049) and this function is called in the [`_next()`](https://github.com/lf-lang/reactor-ts/blob/b5a29738c38222a44b64fc36091e3679e15fe1f7/src/core/reactor.ts#L2074) function. However, there is no comparison like `if (this.isRTISynchronized()...` in the [`lf_next_locked()`](https://github.com/lf-lang/reactor-c/blob/667a5d702cbcd0288328417d8c632e6b5af19c04/core/threaded/reactor_threaded.c#L534) of the reactor-c.

### DIFF
--- a/lingua-franca-ref.txt
+++ b/lingua-franca-ref.txt
@@ -1,1 +1,1 @@
-ts-physical-conn-after-delay
+master
--- a/src/core/federation.ts
+++ b/src/core/federation.ts
@@ -952,6 +952,7 @@ export class FederatedApp extends App {
      * If this variable is true, logical time in this federate
      * cannot advance beyond the time given in the greatest Time Advance Grant
      * sent from the RTI.
+     * FIXME: This concept is very old and not sure why it is needed.
      */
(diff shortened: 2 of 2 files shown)


lhstrh (approved):
LGTM!

PR #150, opened 2023-04-19
Use index variable when access to the `upstreamConnectionDelays` array

byeong-gil:
Until now, `depensOnFedId` has been used for accessing the `upstreamConnectionDelays`.
https://github.com/lf-lang/reactor-ts/blob/5a5f51620f9d6fe0c01076e75ec8e39d6cefc50c/src/core/federation.ts#L1138
However, this can cause an array index out of bound exception.

Thus, in this PR, we use an index variable to access both `dependsOn` and `upstreamConnectionDelays` arrays. Also, we add an exception check to ensure that the length of those two arrays is identical.


src/core/federation.ts
```diff
@@ -1133,14 +1133,16 @@ export class FederatedApp extends App {
         for (let sendsToFedId of config.sendsTo) {
             this.addDownstreamFederate(sendsToFedId);
         }
+        let upstreamConnectionDelaysIndex = 0;
         for (let dependsOnFedId of config.dependsOn) {
             let minOutputConnectionDelay = TimeValue.FOREVER();
-            for (let candidate of config.upstreamConnectionDelays[dependsOnFedId]) {
+            for (let candidate of config.upstreamConnectionDelays[upstreamConnectionDelaysIndex]) {
```
hokeun, line 1139: I don't think there's any boundary check for `upstreamConnectionDelaysIndex` in the current code. How do you make sure that `upstreamConnectionDelaysIndex` does not go out of index bound (e.g., it doesn't go over the size of the array)?

byeong-gil: `upstreamConnectionDelaysIndex` does not exceed the size of the array `config.depensOn`. And the size of `config.dependsOn` and `config.upstreamConnectionDelays` are the same because of [this code](https://github.com/lf-lang/lingua-franca/blob/61f6d678bc2530ce14d2e7697b51ec04d8b5ab52/org.lflang/src/org/lflang/federated/extensions/TSExtension.java#L178).

hokeun (approved):
I updated the code myself to help - please check if it makes sense.
Also, please update the description of this PR to match what this PR actually does before merging this PR.

PR #152, opened 2023-05-19
Install the RTI while testing and Prevent unintended tag advancement

byeong-gil:
In this PR, `ci.yml` is modified to install the RTI while testing federated tests, and unintended tag advancement with the `null` value of `greatestTimeAdvanceGrant` is inhibited. 

Relevant PR
https://github.com/lf-lang/lingua-franca/pull/1752

The RTI should be installed while testing to run federated tests.

Also, the `null` value of `greatestTimeAdvanceGrant` means that there was no TAG message yet. Federates should wait for TAG messages.


hokeun (approved):
Very nice!

PR #153, opened 2023-05-20
Index moved from `src/core` to `src`

lhstrh:
This is done so that `tsc` puts the other `.js` files into `lib/core` even if there are not other files or directories in `src` due to `npm` excluding them based on `.npmignore`.


petervdonovan (approved):
I'm satisfied with this solution!

PR #156, opened 2023-05-28
Fix CI workflow by using the correct URL

axmmisaka:
Merge **either** this or #157.

Right now, the CI/CD uses `"git://github.com/lf-lang/reactor-ts.git#${{ github.head_ref || github.ref_name }}"` as LF reactor runtime. This is not good because PRs from external repositories will fail the CICD test. 

This is one approach to fix it (which I don't like): if it's a PR, get the PR's git URL and set the external reactor to be it.


axmmisaka:
Closing this now, as it appears #157 is preferred.

PR #157, opened 2023-05-28
Fix CI workflow by using locally checked out reactor-ts

axmmisaka:
Merge either this or #156.

Right now, the CI/CD uses `"git://github.com/lf-lang/reactor-ts.git#${{ github.head_ref || github.ref_name }}"` as LF reactor runtime. This is not good because PRs from external repositories will fail the CICD test.

This is one approach to fix it (which I like): checkout the reactor and let LF use that (the checked out) local reactor.

Squash and rebase merge this PR please. Do not just merge because it will look terrible.

I suppose https://github.com/lf-lang/reactor-ts/tree/lhstrh-patch-1 is trying to fix the same thing but sadly `${{ github.repositoryUrl }}` is still always `lf-lang/reactor-ts`......

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,7 +28,7 @@ jobs:
       - name: Check out reactor-ts repository
         uses: actions/checkout@v3
         with: 
-          path: "ts-reactor"
+          path: "reactor-ts"
           fetch-depth: 0
       - name: Read lingua-franca-ref.txt
         run: echo "lingua-franca-ref=$(cat ./reactor-ts/lingua-franca-ref.txt)" >> $GITHUB_ENV
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,10 +28,10 @@ jobs:
       - name: Check out reactor-ts repository
         uses: actions/checkout@v3
         with: 
-          path: "reactor-ts-external"
+          path: "reactor-ts"
           fetch-depth: 0
       - name: Read lingua-franca-ref.txt
-        run: echo "lingua-franca-ref=$(cat ./reactor-ts-external/lingua-franca-ref.txt)" >> $GITHUB_ENV
+        run: echo "lingua-franca-ref=$(cat ./reactor-ts/lingua-franca-ref.txt)" >> $GITHUB_ENV
       - name: Check out lingua-franca repository
         uses: actions/checkout@v3
         with:
@@ -52,4 +52,4 @@ jobs:
         if: ${{ runner.os == 'macOS' || runner.os == 'Linux' }}
       - name: Perform Lingua Franca TypeScript tests
         run: |
-          ./gradlew test --tests org.lflang.tests.runtime.TypeScriptTest.* -Druntime="${{ github.workspace }}/reactor-ts-external"
+          ./gradlew test --tests org.lflang.tests.runtime.TypeScriptTest.* -Druntime="file://${{ github.workspace }}/reactor-ts"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,4 +67,4 @@ jobs:
         if: ${{ runner.os == 'macOS' || runner.os == 'Linux' }}
       - name: Perform Lingua Franca TypeScript tests
         run: |
-          ./gradlew targetTest -Ptarget=TypeScript -Druntime="file:${{ github.workspace }}/reactor-ts"
+          ./gradlew targetTest -Ptarget=TypeScript -Druntime="file:/${{ github.workspace }}/reactor-ts"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,7 +31,7 @@ jobs:
           path: "ts-reactor"
           fetch-depth: 0
       - name: Read lingua-franca-ref.txt
-        run: echo echo "lingua-franca-ref=$(cat ./reactor-ts/lingua-franca-ref.txt)" >> $GITHUB_ENV
+        run: echo "lingua-franca-ref=$(cat ./reactor-ts/lingua-franca-ref.txt)" >> $GITHUB_ENV
       - name: Check out lingua-franca repository
         uses: actions/checkout@v3
         with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,10 +28,10 @@ jobs:
       - name: Check out reactor-ts repository
         uses: actions/checkout@v3
         with: 
-          path: "reactor-ts"
+          path: "reactor-ts-external"
           fetch-depth: 0
       - name: Read lingua-franca-ref.txt
-        run: echo "lingua-franca-ref=$(cat ./reactor-ts/lingua-franca-ref.txt)" >> $GITHUB_ENV
+        run: echo "lingua-franca-ref=$(cat ./reactor-ts-external/lingua-franca-ref.txt)" >> $GITHUB_ENV
       - name: Check out lingua-franca repository
         uses: actions/checkout@v3
         with:
@@ -52,4 +52,4 @@ jobs:
         if: ${{ runner.os == 'macOS' || runner.os == 'Linux' }}
       - name: Perform Lingua Franca TypeScript tests
         run: |
-          ./gradlew test --tests org.lflang.tests.runtime.TypeScriptTest.* -Druntime="${{ github.workspace }}/ts-reactor"
+          ./gradlew test --tests org.lflang.tests.runtime.TypeScriptTest.* -Druntime="${{ github.workspace }}/reactor-ts-external"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,7 +39,6 @@ jobs:
           ref: ${{ env.lingua-franca-ref }}
           fetch-depth: 0
           submodules: true
-          path: "lingua-franca"
       - name: Prepare build environment
         uses: ./.github/actions/prepare-build-env
       - name: Setup Node.js environment
@@ -52,6 +51,5 @@ jobs:
         uses: ./.github/actions/install-rti
         if: ${{ runner.os == 'macOS' || runner.os == 'Linux' }}
       - name: Perform Lingua Franca TypeScript tests
-        working-directory: "./lingua-franca"
         run: |
           ./gradlew test --tests org.lflang.tests.runtime.TypeScriptTest.* -Druntime=${{ github.workspace }}/ts-reactor"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,4 +52,4 @@ jobs:
         if: ${{ runner.os == 'macOS' || runner.os == 'Linux' }}
       - name: Perform Lingua Franca TypeScript tests
         run: |
-          ./gradlew test --tests org.lflang.tests.runtime.TypeScriptTest.* -Druntime=${{ github.workspace }}/ts-reactor"
+          ./gradlew test --tests org.lflang.tests.runtime.TypeScriptTest.* -Druntime="${{ github.workspace }}/ts-reactor"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,8 +25,13 @@ jobs:
   lingua-franca:
     runs-on: ubuntu-latest
     steps:
+      - name: Check out reactor-ts repository
+        uses: actions/checkout@v3
+        with: 
+          path: "ts-reactor"
+          fetch-depth: 0
       - name: Read lingua-franca-ref.txt
-        run: echo "lingua-franca-ref=$(wget -qO- https://raw.githubusercontent.com/lf-lang/reactor-ts/${{ github.head_ref || github.ref_name }}/lingua-franca-ref.txt)" >> $GITHUB_ENV
+        run: echo echo "lingua-franca-ref=$(cat ./reactor-ts/lingua-franca-ref.txt)" >> $GITHUB_ENV
       - name: Check out lingua-franca repository
         uses: actions/checkout@v3
         with:
@@ -34,6 +39,7 @@ jobs:
           ref: ${{ env.lingua-franca-ref }}
           fetch-depth: 0
           submodules: true
+          path: "lingua-franca"
       - name: Prepare build environment
         uses: ./.github/actions/prepare-build-env
       - name: Setup Node.js environment
@@ -46,5 +52,6 @@ jobs:
         uses: ./.github/actions/install-rti
         if: ${{ runner.os == 'macOS' || runner.os == 'Linux' }}
       - name: Perform Lingua Franca TypeScript tests
+        working-directory: "./lingua-franca"
         run: |
-          ./gradlew test --tests org.lflang.tests.runtime.TypeScriptTest.* -Druntime="git://github.com/lf-lang/reactor-ts.git#${{ github.head_ref || github.ref_name }}"
+          ./gradlew test --tests org.lflang.tests.runtime.TypeScriptTest.* -Druntime=${{ github.workspace }}/ts-reactor"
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -25,8 +25,13 @@ jobs:
   lingua-franca:
     runs-on: ubuntu-latest
     steps:
+      - name: Check out reactor-ts repository
+        uses: actions/checkout@v3
+        with: 
+          path: "ts-reactor"
```
lhstrh, line 31:
```suggestion
          path: "reactor-ts"
```

.github/workflows/ci.yml
```diff
@@ -25,8 +25,13 @@ jobs:
...
+          fetch-depth: 0
       - name: Read lingua-franca-ref.txt
-        run: echo "lingua-franca-ref=$(wget -qO- https://raw.githubusercontent.com/lf-lang/reactor-ts/${{ github.head_ref || github.ref_name }}/lingua-franca-ref.txt)" >> $GITHUB_ENV
+        run: echo "lingua-franca-ref=$(cat ./reactor-ts/lingua-franca-ref.txt)" >> $GITHUB_ENV
       - name: Check out lingua-franca repository
```
lhstrh, line 35: Don't we have to designate a `path` for this one as well? The root will already have `reactor-ts` in it, so it won't be empty.

lhstrh: Hm, I guess it's OK, judging by the current run.

axmmisaka: Actually - no, you are right and this is a coincidence, let me change this a bit......

axmmisaka:
I'm thinking naming the checked-out one as `reactor-ts-external`, would that be better?

axmmisaka:
For test result please see https://github.com/axmmisaka/reactor-ts/pull/9.
I think CI/CD of this PR is running against CI/CD config of `master` of this repo which doesn't quite make sense......

.github/workflows/ci.yml
```diff
@@ -25,8 +25,13 @@ jobs:
   lingua-franca:
     runs-on: ubuntu-latest
     steps:
+      - name: Check out reactor-ts repository
+        uses: actions/checkout@v3
+        with: 
+          path: "reactor-ts-external"
```
lhstrh, line 31:
```suggestion
          path: "reactor-ts"
```

.github/workflows/ci.yml
```diff
@@ -47,4 +52,4 @@ jobs:
         if: ${{ runner.os == 'macOS' || runner.os == 'Linux' }}
       - name: Perform Lingua Franca TypeScript tests
         run: |
-          ./gradlew test --tests org.lflang.tests.runtime.TypeScriptTest.* -Druntime="git://github.com/lf-lang/reactor-ts.git#${{ github.head_ref || github.ref_name }}"
+          ./gradlew test --tests org.lflang.tests.runtime.TypeScriptTest.* -Druntime="${{ github.workspace }}/reactor-ts-external"
```
lhstrh, line 55:
```suggestion
          ./gradlew test --tests org.lflang.tests.runtime.TypeScriptTest.* -Druntime="file://${{ github.workspace }}/reactor-ts"
```

.github/workflows/ci.yml
lhstrh: Adding "external" is confusing to me because it's actually using the local version of `reactor-ts`...

.github/workflows/ci.yml
```diff
@@ -25,8 +25,13 @@ jobs:
...
     runs-on: ubuntu-latest
     steps:
+      - name: Check out reactor-ts repository
+        uses: actions/checkout@v3
+        with: 
+          path: "reactor-ts-external"
+          fetch-depth: 0
       - name: Read lingua-franca-ref.txt
-        run: echo "lingua-franca-ref=$(wget -qO- https://raw.githubusercontent.com/lf-lang/reactor-ts/${{ github.head_ref || github.ref_name }}/lingua-franca-ref.txt)" >> $GITHUB_ENV
+        run: echo "lingua-franca-ref=$(cat ./reactor-ts-external/lingua-franca-ref.txt)" >> $GITHUB_ENV
```
lhstrh, line 34:
```suggestion
        run: echo "lingua-franca-ref=$(cat ./reactor-ts/lingua-franca-ref.txt)" >> $GITHUB_ENV
```

axmmisaka:
Oh no, `actions/checkout` will purge the directory and then pull...... This needs some major change I think

axmmisaka:
So now it will first checkout to `./lingua-franca` then files will be moved to `./` (because the main repo's gh action is set up to only be run at pwd). 
Let's see if this works......

lhstrh:
> So now it will first checkout to `./lingua-franca` then files will be moved to `./` (because the main repo's gh action is set up to only be run at pwd). Let's see if this works......
> 
> [axmmisaka#19](https://github.com/axmmisaka/reactor-ts/pull/19)

Hm, I don't think this is necessary. Could can just do two different checkouts, each in their own directory. For the `./gradlew` command, you just specify a current working directory for the step. I don't think we should be copying files around...

lhstrh:
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-set-the-default-shell-and-working-directory

axmmisaka:
> > So now it will first checkout to `./lingua-franca` then files will be moved to `./` (because the main repo's gh action is set up to only be run at pwd). Let's see if this works......
> > [axmmisaka#19](https://github.com/axmmisaka/reactor-ts/pull/19)
> 
> Hm, I don't think this is necessary. Could can just do two different checkouts, each in their own directory. For the `./gradlew` command, you just specify a current working directory for the step. I don't think we should be copying files around...

I [tried it](https://github.com/axmmisaka/reactor-ts/actions/runs/5107917387/jobs/9181266286) and it does not work, reason not being `./gradlew` needs to be run on `pwd` but instead `install-rti` task in `lingua-franca` needs to be run in pwd (because of the way it is designed)
https://github.com/lf-lang/lingua-franca/blob/09151c2ba867de90c0dcb6335f513b54c8ac128a/.github/actions/install-rti/action.yml#L6
and a `working-directory` cannot be specified for `use`. I thought about modifying CICD in the main branch but this is such a corner use case and I do not want to make huge modifications.
https://github.com/lf-lang/lingua-franca/blob/master/.github/actions/install-rti/action.yml

lhstrh:
Oh, wait, you're talking about a different problem. Either way, I think that what you're doing now isn't necessary, but let's see whether it works before we refactor it.

.github/workflows/ci.yml
```diff
@@ -25,15 +25,26 @@ jobs:
...
           fetch-depth: 0
+          path: "lingua-franca"
           submodules: true
+      - name: Move Lingua Franca to default pwd
+        run: |
+          shopt -s dotglob
+          mv -- ./lingua-franca/* ${{ github.workspace }}
+          shopt -u dotglob 
       - name: Prepare build environment
         uses: ./.github/actions/prepare-build-env
```
lhstrh, line 49:
```suggestion
        uses: ./lingua-franca/.github/actions/prepare-build-env
```

.github/workflows/ci.yml
```diff
@@ -47,4 +58,4 @@ jobs:
         if: ${{ runner.os == 'macOS' || runner.os == 'Linux' }}
       - name: Perform Lingua Franca TypeScript tests
         run: |
-          ./gradlew test --tests org.lflang.tests.runtime.TypeScriptTest.* -Druntime="git://github.com/lf-lang/reactor-ts.git#${{ github.head_ref || github.ref_name }}"
+          ./gradlew test --tests org.lflang.tests.runtime.TypeScriptTest.* -Druntime="file://${{ github.workspace }}/reactor-ts"
```
lhstrh, line 61:
```suggestion
          ./gradlew test --tests org.lflang.tests.runtime.TypeScriptTest.* -Druntime="file://${{ github.workspace }}/reactor-ts"
        working-directory: lingua-franca
```

.github/workflows/ci.yml
```diff
@@ -25,15 +25,26 @@ jobs:
...
+        run: echo "lingua-franca-ref=$(cat ./reactor-ts/lingua-franca-ref.txt)" >> $GITHUB_ENV
       - name: Check out lingua-franca repository
         uses: actions/checkout@v3
         with:
           repository: lf-lang/lingua-franca
           ref: ${{ env.lingua-franca-ref }}
           fetch-depth: 0
+          path: "lingua-franca"
           submodules: true
+      - name: Move Lingua Franca to default pwd
```
lhstrh, line 43:
```suggestion
```

.github/workflows/ci.yml
```diff
@@ -25,15 +25,26 @@ jobs:
...
+        run: |
```
lhstrh, line 44:
```suggestion
```

.github/workflows/ci.yml
```diff
@@ -25,15 +25,26 @@ jobs:
...
+          shopt -s dotglob
```
lhstrh, line 45:
```suggestion
```

.github/workflows/ci.yml
```diff
@@ -25,15 +25,26 @@ jobs:
...
+          mv -- ./lingua-franca/* ${{ github.workspace }}
```
lhstrh, line 46:
```suggestion
```

.github/workflows/ci.yml
```diff
@@ -25,15 +25,26 @@ jobs:
...
+          shopt -u dotglob
```
lhstrh, line 47:
```suggestion
```

lhstrh (commented):
How about this? Also, it occurs to me that this workflow should install the RTI, which it's currently not doing.

axmmisaka:
Sorry for the long wait, I tried this before moving to the current non-elegant solution https://github.com/axmmisaka/reactor-ts/commit/b453271f2cb46218d7dfa778ebd28b84e10fe2dc and here is the issue:
While we can invoke `./lingua-franca/.github/actions/install-rti`, the way it is designed is not very compatible: it calls `.github/actions/install-rti/install.sh` and we cannot set its `pwd` here, resulting in file not found error.

lhstrh:
> Sorry for the long wait, I tried this before moving to the current non-elegant solution ([axmmisaka@b453271](https://github.com/axmmisaka/reactor-ts/commit/b453271f2cb46218d7dfa778ebd28b84e10fe2dc)) and here is the issue: While we can invoke `./lingua-franca/.github/actions/install-rti`, the way it is designed is not very compatible: it calls `./.github/actions/install-rti/install.sh` and we cannot set its `pwd` here, resulting in file not found error if the file is instead located in `lingua-frnca/.github/actions/install-rti/install.sh` https://github.com/lf-lang/lingua-franca/blob/master/.github/actions/install-rti/action.yml

I see, but we're currently not calling the `install-rti` action? It's a trivial action, so we don't have to use it. We can also give it an input to specify where to find the sources...

axmmisaka:
> > Sorry for the long wait, I tried this before moving to the current non-elegant solution ([axmmisaka@b453271](https://github.com/axmmisaka/reactor-ts/commit/b453271f2cb46218d7dfa778ebd28b84e10fe2dc)) and here is the issue: While we can invoke `./lingua-franca/.github/actions/install-rti`, the way it is designed is not very compatible: it calls `./.github/actions/install-rti/install.sh` and we cannot set its `pwd` here, resulting in file not found error if the file is instead located in `lingua-frnca/.github/actions/install-rti/install.sh` https://github.com/lf-lang/lingua-franca/blob/master/.github/actions/install-rti/action.yml
> 
> I see, but we're currently not calling the `install-rti` action? It's a trivial action, so we don't have to use it. We can also give it an input to specify where to find the sources...

I think we are referring to two actions defined in `lingua-franca` 
https://github.com/lf-lang/reactor-ts/blob/6d7c64b47dcb08485b4caa3f69e7750ede8fa34b/.github/workflows/ci.yml#L45-L46
https://github.com/lf-lang/reactor-ts/blob/6d7c64b47dcb08485b4caa3f69e7750ede8fa34b/.github/workflows/ci.yml#L37-L38

Both of which were built under the assumption that `lingua-franca` is at pwd.

I think the best way might be stick with the `mv` workaround for now, and edit the lf CI later, because I realised if people use lf, they might need some ways to set up lf in an alternate directory.

axmmisaka:
I'm working on this but lf tests are failing for some reason; I'll take some time to investigate and hopefully resolve by today as #163 is kinda dependent on it...

axmmisaka:
Bug is fixed; before we fix [this issue](https://github.com/lf-lang/reactor-ts/pull/157#issuecomment-1566514249) I think what we have right now is the best solution in terms of compatibility.
I recommend merge this after CICD is finished.

.github/workflows/ci.yml
```diff
@@ -27,26 +27,44 @@ jobs:
...
+          npm run build
       - name: Install pnpm
         run: npm i -g pnpm
       - name: Install RTI
         uses: ./.github/actions/install-rti
         if: ${{ runner.os == 'macOS' || runner.os == 'Linux' }}
       - name: Perform Lingua Franca TypeScript tests
         run: |
-          ./gradlew targetTest -Ptarget=TypeScript -Druntime="git://github.com/lf-lang/reactor-ts.git#${{ github.head_ref || github.ref_name }}"
+          ./gradlew targetTest -Ptarget=TypeScript -Druntime="file:${{ github.workspace }}/reactor-ts"
```
lhstrh, line 70: Not sure about this, but should it be
```suggestion
          ./gradlew targetTest -Ptarget=TypeScript -Druntime="file:/${{ github.workspace }}/reactor-ts"
```

.github/workflows/ci.yml
```diff
@@ -27,26 +27,44 @@ jobs:
...
         with:
           repository: lf-lang/lingua-franca
           ref: ${{ env.lingua-franca-ref }}
           fetch-depth: 0
+          path: "lingua-franca"
           submodules: true
+      - name: Move Lingua Franca to default pwd
+        run: |
+          shopt -s dotglob
+          mv -- ./lingua-franca/* ${{ github.workspace }}
```
lhstrh, line 48: I believe that if you do the checkout of `reactor-ts` without a `path` parameter, it will end up in `${{ github.workspace }}`, and subsequently doing the cloning LF with `path: "lingua-franca"` would put the checkout exactly where you want it? I'm not sure though, but I think all of this should be possible without moving stuff.

axmmisaka: Here are things we want to do:
1. Put `lingua-franca` in `pwd`
2. Put `reactor-ts` in a random directory which we know, other than `pwd` because `lingua-franca` resides there.

Currently, the following procedure is performed:
1. Checkout `reactor-ts` at `pwd/reactor-ts`
2. Checkout `lingua-franca` at `pwd/lingua-franca`
3. Move everything from `pwd/lingua-franca` to `pwd`.

I tried some simplification and they don't work:
1. Not moving `lingua-franca` to `pwd`. [https://github.com/lf-lang/reactor-ts/pull/157#issuecomment-1566514249](This) is the issue.
2. Checkout `reactor-ts` at `pwd/reactor-ts`, then checkout `lingua-franca` at `pwd`. This doesn't work because `actions/checkout@v3` will purge everything in the directory before cloning, so checkout `lingua-franca` at `pwd` will effectively remove everything in `pwd/reactor-ts`.

.github/workflows/ci.yml
axmmisaka: I think `file:/home/lf`, `file://home/lf` `file:///home/lf` all work, but I agree and will use the `//` version.

.github/workflows/ci.yml
lhstrh: OK, let's leave as is because it works 😄

axmmisaka:
I think this is ready to merge? We should squash and merge though, because it contains too many garbage commits which are not needed.

lhstrh (approved):

PR #158, opened 2023-05-30
Add eslint

axmmisaka:
I am using https://github.com/standard/eslint-config-standard-with-typescript/tree/master along with `'eslint:recommended', 'plugin:@typescript-eslint/recommended', 'plugin:@typescript-eslint/recommended-requiring-type-checking'` with semicolons enforced. 

There are looooots of issues that cannot be auto-fixed...... Fixing them one by one by hand.

### DIFF
--- a/src/core/graph.ts
+++ b/src/core/graph.ts
@@ -331,7 +331,7 @@ export class DependencyGraph<T> {
         );
       }
       while (chain.length > 0) {
-        dot += `->' + '"${String(chain.pop())}"`;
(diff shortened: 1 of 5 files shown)


axmmisaka:
Hard lesson: ESLint autofix might break things;

axmmisaka:
@lhstrh I'm not very confident that this  will be passing LF test, as I discovered autofixing introduces a lot of bugs. Would it be more preferable if we add a eslint rule config only for now and not fix anything, and fix code quality issues along the way as we implement other stuff?

Issue #159, opened 2023-06-01
Possibilities of eliminating binding workarounds


axmmisaka:
As I address https://github.com/lf-lang/reactor-ts/pull/158 it appears that there are some nontrivial use of this alias.

https://github.com/lf-lang/reactor-ts/blob/6d7c64b47dcb08485b4caa3f69e7750ede8fa34b/src/core/federation.ts#L359
https://github.com/lf-lang/reactor-ts/blob/6d7c64b47dcb08485b4caa3f69e7750ede8fa34b/src/benchmark/FacilityLocation.ts#L507

I'm a bit curious are there any reason other than scoping that let to such design?

This should be closed as we merge #158 - that is, either we remove these aliases, or remove the rule.

lhstrh:
It's a pattern to keep access to an "outer" `this` after it gets shadowed. Does eslint complain about it? Here is an explanation of the pattern: https://stackoverflow.com/questions/4886632/what-does-var-that-this-mean-in-javascript

axmmisaka:
Yes - it's a rather old design pattern and is discouraged according to eslint (https://typescript-eslint.io/rules/no-this-alias/); it appears that arrow functions (or use `bind`) are encouraged as an alternative. I think this is some re-structuring we could do later.
Additionally, it appears that some of our design which utilised `that` does not suffer from scoping issue at all (if no now closure is created there shouldn't be any scoping issue, iiuc). 

This is what ChatGPT told me......:

> In pre-ES6 JavaScript, it was common to assign `this` to a separate variable, often named `that`, `self`, or `me`, as a workaround to maintain the correct context inside nested functions or callback functions. This was necessary because the inner functions would have their own `this` binding, which could differ from the desired context.
> 
> Here's an example of how the "this binding workaround" was used:
> 
> ```typescript
> function MyClass() {
>   var that = this;
> 
>   this.name = 'MyClass';
> 
>   setTimeout(function() {
>     console.log(that.name); // Accessing the correct context using `that`
>   }, 1000);
> }
> ```
> 
> In this example, `that` is assigned the value of `this`, providing a reference to the desired context within the nested `setTimeout` callback.
> 
> However, with the introduction of arrow functions in ES6, the need for such assignments and workarounds is significantly reduced. Arrow functions do not have their own `this` binding; instead, they inherit the `this` value from the surrounding lexical scope.
> 
> Here's an equivalent version of the previous example using an arrow function:
> 
> ```typescript
> class MyClass {
[...]

PR #160, opened 2023-06-01
Add eslint

axmmisaka:
#158 is a rather complete failure because `eslint` auto fix will introduce a lot of bug.

This PR will gradually introduce fixes, and only auto fix non code-quality issues.

Again - please squash and merge......


package.json
```diff
@@ -14,6 +14,7 @@
         "@types/uuid": "^8.3.4",
         "command-line-args": "^5.1.1",
         "command-line-usage": "^6.1.3",
+        "eslint-config-standard-with-typescript": "^34.0.1",
```
lhstrh, line 17: I think this should be a `devDependency`, no?

lhstrh:
Introducing fixes gradually sounds like a good idea to me. But what will the developer experience be like? Will there be a forest of warnings/errors to wade through?

axmmisaka:
> Introducing fixes gradually sounds like a good idea to me. But what will the developer experience be like? Will there be a forest of warnings/errors to wade through?

Possible; I'll fix as much as possible in this PR today (sorry that I did almost no work yesterday), then discuss in an issue what are the other ones......

__tests__/ActionTrigger.test.ts
```diff
@@ -12,27 +13,27 @@ export class ActionTrigger extends Reactor {
     // This action is never scheduled. It should never be present.
     a2: Action<string> = new Action<string>(this, Origin.logical);
 
-    constructor(parent:Reactor|null) {
+    constructor (parent:Reactor|null) {
```
lhstrh, line 16: All the files have already been formatted, and `eslint` should not introduce new formatting rules. I don't think this looks particularly good, either. Is there a way to ignore or turn off these suggestions for spacing?

src/core/util.ts
```diff
@@ -11,53 +11,53 @@ import ULog from "ulog";
  * @see Log
  */
 export enum LogLevel {
-  ERROR = 1,
-  WARN = 2,
-  INFO = 3,
-  LOG = 4,
-  DEBUG = 5
+    ERROR = 1,
```
lhstrh, line 14: Our spacing is 4, as per prettier config...

__tests__/ActionTrigger.test.ts
axmmisaka: Hmm...... This is weird, because aeb72fb (`eslint-config-prettier`) should've disabled all the rules that conflict with `prettier`. But I will run `prettier` when other changes are done.

src/core/util.ts
axmmisaka: Will disable spacing as well

__tests__/ActionTrigger.test.ts
lhstrh: It looks like this added space is still in the diff. Also, I'd like to be assured that there is no ping-pong game going on between `eslint` and `prettier` where they both want a different thing....

lhstrh: It looks like this would be the culprit: https://eslint.org/docs/latest/rules/space-before-function-paren

src/core/util.ts
lhstrh: I think we would want
```
{
    "indent": ["error", 2]
}
```

__tests__/ActionTrigger.test.ts
axmmisaka: I see what's happening: I think the issue is only happening in `__tests__` and I only fixed relevant issues in `src`. Will push a fix for this now.

> I'd like to be assured that there is no ping-pong game 

Indeed, I realised this was a poor decision on my end, so this came into play:

https://github.com/lf-lang/reactor-ts/blob/174c3177d7ccbc169aefd0ddf27bd5fd19ad8a28/.eslintrc.js#L7

This package basically disables all rules that could potentially conflict with `prettier`, including the one we are discussing.
https://github.com/prettier/eslint-config-prettier/blob/311351ba4d60789fa5d7af5bacd0e6a8dc3c7224/index.js#L99

src/core/util.ts
axmmisaka: I think the preferred way is to disable the rule (like what's happening now) and let prettier handle this.

__tests__/ActionTrigger.test.ts
lhstrh: That sounds right, but the "space-before-function-paren" rule seems to have still been applied. The line that this discussion is anchored to still has it, for instance.

lhstrh (commented):
I generally like the changes made, except for the "space-before-function-paren" rule.

lhstrh (approved):
Looks great! Has all the noise that eslint makes been addressed?

axmmisaka:
Yes and no - I made sure `eslint` won't complain, but some are simply suppressed by `eslint-disable-next-line` and added as TODO items.
Here are some examples:
1. `@typescript-eslint/no-this-alias` is suppressed entirely; it is addressed in #159 
2. Some Java style code that does not live well with ts, for example, empty abstract class and classes with only static methods
3. https://github.com/lf-lang/reactor-ts/blob/6db9fb0cf90be7c6acd75dbde2e2fcee8f961e61/src/core/port.ts#L90-L99 ; if this is changed the whole `reactor.ts` will explode
4. We have a lot of use of `any` and maybe there could be alternatives

Yet I think these changes will involve complex refactoring and should be addressed in individual issues

PR #161, opened 2023-06-03
Do not run CI on non-master branches

axmmisaka:
For non-master branches, they are probably all created for some PR and a duplication CI problem arises when a PR is created. It makes more sense to run CI only on `master` and PRs.

image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,6 +2,8 @@ name: CI
 
 on:
   push:
+    branches:
+      - master
   pull_request:
   # Also allow running this workflow manually from the Actions tab.
   workflow_dispatch:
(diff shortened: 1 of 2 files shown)


lhstrh:
The CI configuration changed on `master`, which is why this doesn't "just work" as it normally would...

axmmisaka:
I see; that definitely needs to be addressed and I'll look into it after #160 is ready to be reviewed......

axmmisaka:
> Note: the CI configuration changed on `master` of the `lingua-franca` repo. I noticed that you changed `lingua-franca-ref.txt` in #160, but that will require reworking `ci.yml` a little bit. I'm OK with only running CI on `master` and PRs through.

I think it works now? Thanks for the fix!

lhstrh (approved):

PR #165, opened 2023-06-09
Remove alias to this and remove relevant rule in eslint

### LINKED ISSUE #159: Possibilities of eliminating binding workarounds

axmmisaka:
This addresses and should close #159.

As discussed, most of the usage of `this` results from es `function() {}` being unable to capture `this` of its parent as it itself has its own `this` (what a tongue twister......); in our code, most of the uses of `this`-alias is completely unnecessary, and some results in the above-mentioned issue which can be solved by changing it to anonymous arrow function `() => {}` which does not have its own `this` and will use its parent's `this`.

If tests are not passing, I suppose the issue would be that some `this` aliases are passed into its grandchildren; but vscode is not warning me about the existence of relevant issue.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -15,8 +15,6 @@ module.exports = {
   root: true,
   rules: {
     "@typescript-eslint/quotes": ["error", "double"],
-    // This is to be addressed separately later
-    "@typescript-eslint/no-this-alias": "off",
     "@typescript-eslint/restrict-template-expressions": "off",
     "@typescript-eslint/naming-convention": [
       "warn",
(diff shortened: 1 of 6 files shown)


lhstrh (approved):
This looks good to me! Thanks for the cleanup, @axmmisaka!

Issue #167, opened 2023-06-19
Issues with Present and Remedies


axmmisaka:
I had a discussion with @lhstrh lately about the design and discussed the elimination of `Present`. It consists two parts:

1. Removal of `null` and its alternative
For this, we decided to remove `null` because it adds some confusion in typing and type-checking. The reason `null` was introduced was to denote "Present without any payload". The way to go, in our opinion, might be using an empty string `""` or an empty TS Object (`{}`). I prefer the second one, as it has its own type: `Record<string, never>`.

2. Removal of `Present`?
Following 1, I experimented and will propose 2 ways to move forward:

a. Do not remove `Present`, only replace `null`
https://github.com/lf-lang/reactor-ts/tree/axmmisaka/remove-present-null
There might be good reason to keep `Present` which I'm not entirely sure; yet, if we were to keep it, this is what it will look like

https://github.com/lf-lang/reactor-ts/compare/master...axmmisaka/remove-present-null#diff-80ebcd3dc48e85c12b2c8484c5355662ab98ceea358660eb3af536c584f6fa44R85-R93

https://github.com/lf-lang/reactor-ts/compare/master...axmmisaka/remove-present-null#diff-3827285505d863646ccfca9d3df43654ced3c24d054b1698b142c4c50e5b2a8eL2520-R2520

b. Remove `Present` and replace with `unknown`
https://github.com/lf-lang/reactor-ts/tree/axmmisaka/remove-present-replace-unknown
If we were to remove `Present`, the best way to replace it is to make it being able to hold any value. In functions, previous generic types are used, but `extends Present` are deleted to make them implicitly extend `unknown`.

There might, of course, be better design ideas which we might need further discussion, but these are what I have now......

axmmisaka:
There is another way which could be better, that is defining `Present` to be `NonNullable<unknown>`.

axmmisaka:
Closed as #168 is merged

PR #168, opened 2023-06-19
Remove `Present` and replace with `unknown`; remove Args and Trigger and use TS Tuple

axmmisaka:
See https://github.com/lf-lang/reactor-ts/issues/167

TODOs:
- [x] Make accompanying lingua-franca PR (https://github.com/lf-lang/lingua-franca/pull/1857) in that eliminates `null`
- [ ] ~~Change `null` to be `{} satisfies Record<string, never>`, check https://github.com/lf-lang/reactor-ts/tree/axmmisaka/remove-present-replace-unknown~~
- [ ] ~~Enforce `null` check~~

For reference:
> I discovered what happened; such a type gymnastics moment
> It is because typescript will infer `T extends unknown` to be `null` and it is still valid;
> If we remove `Present` and still want to check against `null`, we make all `T` to be `T extends NonNullable<unknown>`
> Or, we can just define `Present` as such...... (i.e. `type Present = NonNullable<unknown>`)?
> 
> But I don't think checking is entirely necessary, considering all tests are passing...... And we might be checking null already somewhere in the code possibly?

### DIFF
--- a/__tests__/ActionTrigger.test.ts
+++ b/__tests__/ActionTrigger.test.ts
@@ -1,7 +1,6 @@
 import type {Sched} from "../src/core/internal";
 import {
   App,
-  Tuple,
   Origin,
   TimeValue,
   Reactor,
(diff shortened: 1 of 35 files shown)


lhstrh (approved):
Looks great!

PR #170, opened 2023-06-22
Refactor graph package (addresses #166)

### LINKED ISSUE #166: Graph Issues
Currently, we encode dependencies using a dependency graph. We are considering inverting the polarity of the edges, effectively turning it into a precedence graph. Before going ahead with this, we need to:
 - [ ] understand all current uses of the graph
 - [ ] understand whether changing the polarity will have algorithmic disadvantages
 - [ ] come up with an API that we like
 - [ ] stick to clear terminology across the code base
 - [ ] Migrate DOT notation to `mermaid.js`
 - [ ] ......

axmmisaka:
This PR fixes #166. 
- [x] Rename APIs so they less confusing
- [x] Remove unused or redundant APIs (removed `addBackEdges`/`addEdges`)
- [x] Move `toString` to use `mermaid.js` representation
- [x] Rewrite search functions to use iteration, and use more efficient algorithm (rewrote `hasCycle`; the one in `fromDependencyGraph` will incur some overhead so omitted)
- [x] Refactor improper use of graph APIs in `Reactor`

### DIFF
--- a/__tests__/dependencies.ts
+++ b/__tests__/dependencies.ts
@@ -64,7 +64,7 @@ describe("Manually constructed simple precedence graphs", () => {
   graph.addNode(nodes[0]);
 
   it("graph equality", () => {
-    expect([...graph.nodes()]).toEqual(nodes);
+    expect([...graph.getNodes()]).toEqual(nodes);
   });
 });
 
@@ -142,7 +142,8 @@ describe("Manually constructed precedence graphs", () => {
 
   it("add node 7, make 3 dependent on it", () => {
     graph.addNode(nodes[6]);
-    graph.addEdges(nodes[2], new Set([nodes[6], nodes[3]]));
+    graph.addEdge(nodes[2], nodes[6]);
+    graph.addEdge(nodes[2], nodes[3]);
     expect(graph.size()[0]).toEqual(6); // V
     expect(graph.size()[1]).toEqual(4); // E
     Log.global.debug(graph.toString());
--- a/__tests__/dependencies.ts
+++ b/__tests__/dependencies.ts
@@ -1,4 +1,4 @@
-import type {Priority, Sortable} from "../src/core/internal";
+import type {Reaction, Priority, Sortable} from "../src/core/internal";
 import {
   Reactor,
   App,
@@ -8,8 +8,7 @@ import {
   SortablePrecedenceGraph,
   PrioritySet,
   Log,
-  StringUtil,
-  Reaction
+  StringUtil
 } from "../src/core/internal";
 
 // Log.setGlobalLevel(Log.levels.DEBUG);
@@ -57,7 +56,7 @@ class CNode<T> implements Sortable<Priority> {
 }
 
 describe("Manually constructed simple precedence graphs", () => {
-  var graph = new SortablePrecedenceGraph<Sortable<Priority>>(Reaction<unknown>);
+  var graph = new SortablePrecedenceGraph<Sortable<Priority>>();
   var reactor = new SR(new App());
 
   var nodes = reactor.getNodes();
@@ -76,7 +75,7 @@ describe("Test for corner cases", () => {
 });
 
 describe("Manually constructed precedence graphs", () => {
-  var graph = new SortablePrecedenceGraph<Sortable<Priority>>(Reaction<unknown>);
+  var graph = new SortablePrecedenceGraph<Sortable<Priority>>();
   var reactor = new R(new App());
 
   var nodes = reactor.getNodes();
@@ -205,7 +204,7 @@ describe("Manually constructed precedence graphs", () => {
 });
 
 describe("ReactionQ", () => {
-  var graph = new SortablePrecedenceGraph<Reaction<unknown>>(Reaction<unknown>);
+  var graph = new SortablePrecedenceGraph<Reaction<unknown>>();
   var reactor = new R(new App());
 
   var nodes = reactor.getNodes();
--- a/__tests__/dependencies.ts
+++ b/__tests__/dependencies.ts
@@ -1,4 +1,4 @@
-import type {Reaction, Priority, Sortable} from "../src/core/internal";
+import type {Priority, Sortable} from "../src/core/internal";
 import {
   Reactor,
   App,
@@ -8,7 +8,8 @@ import {
   SortablePrecedenceGraph,
   PrioritySet,
   Log,
-  StringUtil
+  StringUtil,
+  Reaction
 } from "../src/core/internal";
 
 // Log.setGlobalLevel(Log.levels.DEBUG);
@@ -56,7 +57,7 @@ class CNode<T> implements Sortable<Priority> {
 }
 
 describe("Manually constructed simple precedence graphs", () => {
-  var graph = new SortablePrecedenceGraph<Sortable<Priority>>();
+  var graph = new SortablePrecedenceGraph<Sortable<Priority>>(Reaction<unknown>);
   var reactor = new SR(new App());
 
   var nodes = reactor.getNodes();
@@ -69,13 +70,13 @@ describe("Manually constructed simple precedence graphs", () => {
 });
 
 describe("Test for corner cases", () => {
-  var graph = new SortablePrecedenceGraph<Sortable<Priority>>();
+  var graph = new SortablePrecedenceGraph<Sortable<Priority>>(CNode<Priority>);
   const node: Sortable<Priority> = new CNode<Priority>();
   graph.addEdge(new CNode<Priority>(), node);
 });
 
 describe("Manually constructed precedence graphs", () => {
-  var graph = new SortablePrecedenceGraph<Sortable<Priority>>();
+  var graph = new SortablePrecedenceGraph<Sortable<Priority>>(Reaction<unknown>);
   var reactor = new R(new App());
 
   var nodes = reactor.getNodes();
@@ -204,7 +205,7 @@ describe("Manually constructed precedence graphs", () => {
 });
 
 describe("ReactionQ", () => {
-  var graph = new SortablePrecedenceGraph<Reaction<unknown>>();
+  var graph = new SortablePrecedenceGraph<Reaction<unknown>>(Reaction<unknown>);
   var reactor = new R(new App());
 
   var nodes = reactor.getNodes();
--- a/__tests__/dependencies.ts
+++ b/__tests__/dependencies.ts
@@ -96,11 +96,21 @@ describe("Manually constructed precedence graphs", () => {
     expect(graph.size()[0]).toEqual(6); // V
     expect(graph.size()[1]).toEqual(7); // E
     expect(graph.toString()).toBe(
-      StringUtil.dontIndent`digraph G {
-            "app.R[R0]"->"app.R[R1]"->"app.R[R4]"->"app.R[R3]"->"app.R[R5]";
-            "app.R[R0]"->"app.R[R4]";
-            "app.R[R1]"->"app.R[R2]"->"app.R[R3]";
-            }`
+      StringUtil.dontIndent
+      `graph
+        0["app.R[R3]"]
+        1["app.R[R5]"]
+        2["app.R[R4]"]
+        3["app.R[R2]"]
+        4["app.R[R1]"]
+        5["app.R[R0]"]
+        1 --> 0
+        0 --> 2
+        0 --> 3
+        3 --> 4
+        2 --> 4
+        4 --> 5
+        2 --> 5`
     );
   });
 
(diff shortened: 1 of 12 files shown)


src/core/graph.ts
```diff
@@ -296,14 +276,57 @@ export class DependencyGraph<T> {
...
   }
 
+  toString: () => string = () => this.toMermaidRepresentation();
+
+  /**
+   * Return a representation that conforms with the syntax of mermaid.js
+   * @param edgesWithIssue Edges in the **dependency** graph that causes issues
+   * to the execution. A set containing arrays with [effect, origin].
+   */
+  toMermaidRepresentation(edgesWithIssue?: Set<[T, T]>): string {
```
axmmisaka, line 290: `edgesWithIssue` is for situations that might arise here: 

https://github.com/lf-lang/reactor-ts/blob/c01a31c7d400bd5a31e9707e4293cfa791d886b7/src/core/reactor.ts#L1130-L1145

If the introduction of edge(s) makes the graph problematic, such as introducing `app.R[R2] -> app.R[R5]` in this graph (this is the precedence graph; in reality a dependency, i.e. edge, from R5 to R2 is introduced):

```mermaid
graph
    0["app.R[R3]"]
    1["app.R[R5]"]
    2["app.R[R4]"]
    3["app.R[R2]"]
    4["app.R[R0]"]
    5["app.R[R6]"]
    1 --> 0
    0 --> 3
    5 --> 3
    2 --> 4
```

Then it will become this indicating that the newly-added edge is problematic.
```mermaid
graph
    0["app.R[R3]"]
    1["app.R[R5]"]
    2["app.R[R4]"]
    3["app.R[R2]"]
    4["app.R[R0]"]
    5["app.R[R6]"]
    1 --> 0
    0 --> 3
    5 --> 3
    2 --> 4
    3 --x 1
```

src/core/graph.ts
```diff
@@ -150,17 +150,11 @@ export class DependencyGraph<T> {
...
-      return nodes;
-    } else {
-      return new Set<T>();
-    }
+  getOriginsOfEffect(node: T): Set<T> {
+    return this.adjacencyMap.get(node) ?? new Set<T>();
   }
 
-  getBackEdges(node: T): Set<T> {
+  getEffectsOfOrigin(node: T): Set<T> {
```
axmmisaka, line 157: Using `origin` and `effect` very effectively eliminated the confusion while using the API; yet it is a bit confusing when writing functions within the class that are dependent upon these functions

__tests__/dependencies.ts
```diff
@@ -96,11 +96,21 @@ describe("Manually constructed precedence graphs", () => {
     expect(graph.size()[0]).toEqual(6); // V
     expect(graph.size()[1]).toEqual(7); // E
     expect(graph.toString()).toBe(
-      StringUtil.dontIndent`digraph G {
-            "app.R[R0]"->"app.R[R1]"->"app.R[R4]"->"app.R[R3]"->"app.R[R5]";
-            "app.R[R0]"->"app.R[R4]";
-            "app.R[R1]"->"app.R[R2]"->"app.R[R3]";
-            }`
+      StringUtil.dontIndent
```
axmmisaka, line 99: These tests might need to be changed, as the representation of `mermaid.js` here does not have to conform to a specific ordering.

src/core/graph.ts
```diff
@@ -424,6 +447,43 @@ export class DependencyGraph<T> {
 export class SortableDependencyGraph<
   T extends Sortable<number>
 > extends DependencyGraph<T> {
+  static fromDependencyGraph<R, T extends Sortable<number>>(
```
axmmisaka, line 450: This practice is very Java-y, but I don't think using a constructor would be a good idea as ts doesn't allow overloading so that will cause some headache.

src/core/graph.ts
lhstrh: Maybe using upstream/downstream would be even better? E.g., `getUpstreamNeighbors` or `getUpstreamNodes` or simply `getUpstream`.

lhstrh: We could also use source/sink terminology...

src/core/graph.ts
```diff
@@ -187,40 +181,47 @@ export class DependencyGraph<T> {
...
-    }
+    };
     search(effect);
     reachable.delete(effect);
 
     return reachable;
   }
 
+  /**
+   * This uses DFS with iteration to check for back edges in the graph.
```
lhstrh, line 195: :+1:

src/core/graph.ts
lhstrh: Using Mermaid is such an improvement!

__tests__/dependencies.ts
lhstrh: What do you mean, exactly? Is the function that produces the output printing things in an arbitrary order?

src/core/graph.ts
```diff
@@ -410,10 +433,10 @@ export class DependencyGraph<T> {
   // have a graph without any leaf nodes.
   // As a result, starting a graph search only from leaf nodes in a
   // cyclic graph, will not necessarily traverse the entire graph.
-  public leafNodes(): Set<T> {
-    const leafs = new Set<T>(this.nodes());
-    for (const node of this.nodes()) {
-      for (const dep of this.getEdges(node)) {
+  public pureEffectNodes(): Set<T> {
```
lhstrh, line 436: I'm not sure I like this change. While anyone who has dealt with graphs would know what a "leaf" or "root" node is, term "pure" is rather elusive.

__tests__/dependencies.ts
axmmisaka: It is not printing things in arbitrary order. Nevertheless, our test should not expect it to be printing things in some certain order

src/core/graph.ts
axmmisaka: The issue with `leaf` or `root` terminology is that it is a bit confusing to know whether it is referring to leaf with respect to the underlying data structure or the direction of data flow.
I do think this naming is not as satisfactory and I'll think about that names could be unambiguous and descriptive.

__tests__/dependencies.ts
lhstrh: Why not?

axmmisaka: It relies on 1. the fact that JS sets maintain insertion order, and 2. we are looping in that order. If there is an implementation that does not loop in such an order, or if 1 does not hold for `nodejs` anymore (which is unlikely), then the tests will break while in reality it shouldn't. I don't think that this is worth putting too much effort on this as of now, but we (probably) should bear this in mind.

lhstrh: We are testing our own implementation, and if it breaks because of changes in the underlying data structures (which seems unlikely to happen), then this leading to test failures would be fine.

src/core/graph.ts
axmmisaka: After a second thought, I think it would be a good idea to do so; as long as we consistently expose API names as a precedence graph I think it would be fine? We would need to document these API calls though, so that people are not confused.

axmmisaka:
I feel that this is a bit bike-shedding, yet this conversation is also crucial in terms of making the API more understandable.

| Original  | Data Flow/Precedence Graph Perspective  |  Dependency Graph Perspective  | Alias |
|---|---|---|---|
| `getEdges`  | `getUpstreamNeighbor` |  `getChildren`  | `getChildren` + `getUpstreamNeighbor` |
| `getBackEdges`  |  `getDownstreamNeighbor` | `getParents`  | `getParents`  + `getDownstreamNeighbor` |
| `rootNodes`  | `sourceNodes`/`nodesWithNoOrigin`?  | `sinkNodes`  | `sourceDataNodes`+`sinkGraphNodes` |
| `leafNodes`  | `sinkNodes`/`nodesWithNoEffect`?  | `sourceNodes`  | `sinkDataNodes`+`sourceGraphNodes` |

Let's just pick one and move on; I don't want to use graph terminology solely because it confuses people with the data flow/dependency graph issue (since the class is called `DependencyGraph`, it might be natural to assume no assumption on the data flow; although we use the graph solely to model the data flow between reactions/ports of reactors).

edwardalee:
In real life, children depend on parents. Here, it is reversed.

I never liked the dependency graph formulation.  Having arrows that point in the opposite direction of data flow confuses me.

edwardalee:
In real life, children depend on parents. Here, it is reversed.

I never liked the dependency graph formulation.  Having arrows that point in the opposite direction of data flow confuses me.

axmmisaka:
> In real life, children depend on parents. Here, it is reversed.
> 
> I never liked the dependency graph formulation. Having arrows that point in the opposite direction of data flow confuses me.

The reason we use and keep using a dependency graph, imo, is to keep some algorithmic advantage (since we trace to upstream reactions more often)
I think the best way to go is to abstract out the existence of a dependency graph. We expose an abstract data model that represents reaction hierarchy without mentioning dependency or precedence, and name the APIs with natural data flow

lhstrh:
> I never liked the dependency graph formulation. Having arrows that point in the opposite direction of data flow confuses me.

Same for me. The refactoring that @axmmisaka is working on is meant to address this.

axmmisaka:
> > I never liked the dependency graph formulation. Having arrows that point in the opposite direction of data flow confuses me.
> 
> Same for me. The refactoring that @axmmisaka is working on is meant to address this.

I think it would be great if we could figure out a balanced plan during today's meeting, and execute it by today. We might be spending too much time on this and the issue is not very technical......

If we were to keep the underlying data structure to be a dependency graph, maybe aliasing would be the best idea to prevent headache......

src/core/graph.ts
```diff
@@ -144,23 +42,32 @@ export class DependencyGraph<T> {
...
-    } else {
-      return new Set<T>();
-    }
+  
+  /**
+   * Return the set of all incoming nodes of the given node.
+   * Incoming nodes are immediate upstream neighbors.
+   * @param node The node to retrieve the incoming nodes of.
+   */
+  getInNodes(node: T): Set<T> {
```
axmmisaka, line 61: Personally I still prefer `upstream` and `downstream`, because "in node/"in vertex" doesn't seem like a terminology that is very commonly used in terms of graphs. `parent` and `children` works as well

src/core/reaction.ts
```diff
@@ -105,7 +105,7 @@ export class Reaction<T>
    * @param another Reaction to compare this reaction's priority against.
    */
   hasPriorityOver(another: PrioritySetElement<Priority> | undefined): boolean {
-    if (another != null && this.getPriority() < another.getPriority()) {
+    if (another !== undefined && this.getPriority() < another.getPriority()) {
```
axmmisaka, line 108: We should revert it back, it appears that this is a caching issue and could be fixed by re-running `npx esline` or restarting TS server

src/core/reactor.ts
```diff
@@ -2587,36 +2581,12 @@ export class App extends Reactor {
...
-    }
-
-    for (const leaf of leafs) {
-      if (leaf instanceof Reaction) {
-        collapsed.addNode(leaf);
-        search(leaf, apg.getEdges(leaf));
-        visited.clear();
-      }
-    }
+    const collapsed = SortablePrecedenceGraph.fromPrecedenceGraph(
```
axmmisaka, line 2586: Same here, the issue we encountered during our meeting today is reproducible in VSCode but eslint doesn't complain

src/core/graph.ts
lhstrh: The problem with `upstream` and `downstream` fails to specify how far up or downstream. I think the definitions given in the comments are pretty clear, no? Incoming nodes are immediate upstream neighbors. Outgoing nodes are immediate downstream neighbors.

src/core/reaction.ts
lhstrh: OK. Is the original code correct and the change wrong?

src/core/graph.ts
axmmisaka: > The problem is that upstream and downstream fail to specify how far up or downstream
I agree, that might be an issue

> I think the definitions given in the comments are pretty clear, no?
Yes, yet the issue is people might need to read the comments to know what exactly it is. Typically "incoming" and "outgoing" are adjectives to "edges" (e.g. "incoming edges"); wrt this use I think the term parent/ancestors and children/descendants are more commonly used and easily understandable.

But since these methods are well-documented I think people will know what they are.

src/core/reaction.ts
axmmisaka: The original code is correct yet the change is not wrong. It's just (I feel) a common practice to always use `!= null` instead of `!== undefined`, even if the variable is non-nullable.

src/core/graph.ts
```diff
@@ -392,45 +316,99 @@ export class DependencyGraph<T> {
...
+> extends PrecedenceGraph<T> {
+  /**
+   * Create a sortable precedence graph. If a type and precedence graph are given,
+   * then remove all nodes that are not of the given type in a way the preserves
+   * the original lineage.
+   * @param type A type that extends T.
+   * @param pg A precedence graph.
+   */
+  constructor(
+    type?: new (...args: never[]) => T,
```
lhstrh, line 361: @axmmisaka, I wasn't 100% sure about this change. It used to be `any[]` and this appears to work well without raising any eslint problems...

lhstrh (approved):
I think we've arrived at a point to where this looks ready to merge...

src/core/graph.ts
axmmisaka: I guess this is fine and intended, since we *never* assign to the arguments and this is meant to be a placeholder for "any argument"

axmmisaka:
As I add tests, I discovered that there are some small bugs; I'll fix them, then this should be ready to be merged......

axmmisaka:
I fixed a bug in the mermaid representation's `edgesWithIssue`, added tests to test `edgesWithIssue`, and ran `prettier` on relevant files. I think this should be good to go, other than more comments on `queue.ts` which we could and should address later.

lhstrh (approved):
:rocket:

PR #174, opened 2023-07-01
Remove Args and Trigger

axmmisaka:
Majority of this PR was made by @lhstrh in #172. This is a rebased version of it so that we might be able to have more discussion on the elimination of `Present` while merging this.


lhstrh (approved):
This looks great! I wonder why there is a `ReactorArgs<T>` definition in `bank.ts`, which doesn't seem to belong there that was not introduced in this PR.

axmmisaka:
Closed, as this is included in #168 and it is merged.

PR #186, opened 2023-07-06
`NetworkSender` and `NetworkSender` classes for new federated execution scheme

byeong-gil:
This PR is a companion of https://github.com/lf-lang/lingua-franca/pull/1684 and provides changes to the infrastructure that supports federated execution.

Specifically, it adds the `NetworkReceiver` and `NetworkSender` classes to make compatibility with the [LF PR#1684](https://github.com/lf-lang/lingua-franca/pull/1684). Also, the reactions that send out "port absent" messages, are now dealt with differently.

Note that this PR has no ability to pass 'failing tests' with a cyclic dependency. MLAA-based reaction execution and dependencies implied by TPO levels will be addressed in this [reactor-ts PR#198](https://github.com/lf-lang/reactor-ts/pull/198).

### DIFF
--- a/__tests__/connection.test.ts
+++ b/__tests__/connection.test.ts
@@ -39,12 +39,6 @@ describe("Check canConnect", () => {
           true
         );
       });
-
-      it("canConnect failure", () => {
-        expect(this.canConnect(this.destination.in, this.source.out)).toBe(
-          false
-        );
-      });
     }
   }
   var testApp = new TestApp();
--- a/__tests__/connection.test.ts
+++ b/__tests__/connection.test.ts
@@ -39,6 +39,12 @@ describe("Check canConnect", () => {
           true
         );
       });
+
+      it("canConnect failure", () => {
+        expect(this.canConnect(this.destination.in, this.source.out)).toBe(
+          false
+        );
+      });
     }
   }
   var testApp = new TestApp();
--- a/__tests__/hierarchy.ts
+++ b/__tests__/hierarchy.ts
@@ -54,9 +54,6 @@ describe("Container to Contained", () => {
     expect(
       app.container.canConnect(app.container.b, app.container.contained.a)
     ).toBe(false);
-    expect(
-      app.container.canConnect(app.container.contained.a, app.container.b)
-    ).toBe(false);
 
     expect(
       app.container.canConnect(app.container.b, app.container.contained.b)
--- a/__tests__/hierarchy.ts
+++ b/__tests__/hierarchy.ts
@@ -48,9 +48,6 @@ describe("Container to Contained", () => {
     ).toBe(true);
 
     expect(
-      app.container.canConnect(app.container.a, app.container.contained.b)
-    ).toBe(false);
-    expect(
       app.container.canConnect(app.container.contained.b, app.container.a)
     ).toBe(false);
 
--- a/__tests__/hierarchy.ts
+++ b/__tests__/hierarchy.ts
@@ -42,12 +42,6 @@ describe("Container to Contained", () => {
     ).toBe(false);
     expect(
       app.container.canConnect(
-        app.container.contained.a,
-        app.container.contained.b
-      )
-    ).toBe(false);
-    expect(
-      app.container.canConnect(
         app.container.contained.b,
         app.container.contained.a
       )
--- a/__tests__/hierarchy.ts
+++ b/__tests__/hierarchy.ts
@@ -42,17 +42,35 @@ describe("Container to Contained", () => {
     ).toBe(false);
     expect(
       app.container.canConnect(
+        app.container.a,
+        app.container.b
+      )
+    ).toBe(true);
+    expect(
+      app.container.canConnect(
+        app.container.contained.a,
+        app.container.contained.b
+      )
+    ).toBe(false);
+    expect(
+      app.container.canConnect(
         app.container.contained.b,
         app.container.contained.a
       )
     ).toBe(true);
 
     expect(
+      app.container.canConnect(app.container.a, app.container.contained.b)
+    ).toBe(false);
+    expect(
       app.container.canConnect(app.container.contained.b, app.container.a)
     ).toBe(false);
 
     expect(
       app.container.canConnect(app.container.b, app.container.contained.a)
+    ).toBe(false);
+    expect(
+      app.container.canConnect(app.container.contained.a, app.container.b)
     ).toBe(false);
 
     expect(
@@ -65,6 +83,9 @@ describe("Container to Contained", () => {
     expect(app.container.canConnect(app.container.contained.a, app.foo.a)).toBe(
       false
     );
+    expect(app.container.canConnect(app.container.contained.a, app.foo.b)).toBe(
+      false
+    );
     expect(app.container.canConnect(app.foo.a, app.container.contained.a)).toBe(
       false
     );
@@ -72,6 +93,7 @@ describe("Container to Contained", () => {
       false
     );
 
+    expect(app.container.canConnect(app.foo.a, app.container.b)).toBe(false);
     expect(app.container.canConnect(app.foo.a, app.container.a)).toBe(false);
 
     // expect(app.container.contained).toBeDefined();
--- a/__tests__/hierarchy.ts
+++ b/__tests__/hierarchy.ts
@@ -65,9 +65,6 @@ describe("Container to Contained", () => {
     expect(app.container.canConnect(app.container.contained.a, app.foo.a)).toBe(
       false
     );
-    expect(app.container.canConnect(app.container.contained.a, app.foo.b)).toBe(
-      false
-    );
     expect(app.container.canConnect(app.foo.a, app.container.contained.a)).toBe(
       false
     );
@@ -75,7 +72,6 @@ describe("Container to Contained", () => {
       false
     );
 
-    expect(app.container.canConnect(app.foo.a, app.container.b)).toBe(false);
     expect(app.container.canConnect(app.foo.a, app.container.a)).toBe(false);
 
     // expect(app.container.contained).toBeDefined();
--- a/src/core/federation.ts
+++ b/src/core/federation.ts
@@ -295,7 +295,7 @@ export class NetworkSender extends Reactor {
    * The last reaction of a NetworkSender reactor is 'portAbsentReactor'.
    * @returns portAbsentReactor of this NetworkSender reactor
    */
-  public getLastReactioOrMutation(): Reaction<Variable[]> | undefined {
+  public getLastReactionOrMutation(): Reaction<Variable[]> | undefined {
     return this._getLastReactionOrMutation();
   }
 }
@@ -309,21 +309,6 @@ export class NetworkReceiver<T> extends Reactor {
    */
   private networkInputAction: FederatePortAction<T> | undefined;
 
-  // The port ID of networkInputAction.
-  private readonly portID: number;
-
-  constructor(parent: Reactor, portID: number) {
-    super(parent);
-    this.portID = portID;
-  }
-
-  /**
-   * Getter for portID of this NetworkReactor.
-   */
-  public getPortID(): number | undefined {
-    return this.portID;
-  }
-
   /**
    * Register a federate port's action with the network receiver.
    * @param networkInputAction The federate port's action for registration.
@@ -336,18 +321,11 @@ export class NetworkReceiver<T> extends Reactor {
 
   /**
    * Handle a message being received from the RTI.
-   * @param portID The destination port ID of the message.
    * @param value The payload of the message.
    */
-  public handleMessage(portID: number, value: T): void {
+  public handleMessage(value: T): void {
     // Schedule this federate port's action.
     // This message is untimed, so schedule it immediately.
-    if (portID !== this.portID) {
-      this.util.reportError(
-        "FederatedApp attempts to pass the tagged message to the wrong port ID"
-      );
-      return;
-    }
     if (this.networkInputAction !== undefined) {
       this.networkInputAction
         .asSchedulable(this._getKey(this.networkInputAction))
@@ -360,7 +338,7 @@ export class NetworkReceiver<T> extends Reactor {
    * @param portID The destination port ID of the message.
    * @param value The payload of the message.
    */
-  public handleTimedMessage(portID: number, value: T, intendedTag: Tag): void {
+  public handleTimedMessage(value: T, intendedTag: Tag): void {
     // Schedule this federate port's action.
 
     /**
@@ -382,12 +360,6 @@ export class NetworkReceiver<T> extends Reactor {
 
     // FIXME: implement decentralized control.
 
-    if (portID !== this.portID) {
-      this.util.reportError(
-        "FederatedApp attempts to pass the tagged message to the wrong port ID"
-      );
-      return;
-    }
     if (this.networkInputAction !== undefined) {
       if (this.networkInputAction.origin === Origin.logical) {
         this.networkInputAction
@@ -1215,7 +1187,10 @@ export class FederatedApp extends App {
   /**
    * An array of network receivers
    */
-  private readonly networkReceivers: Array<NetworkReceiver<unknown>> = [];
+  private readonly networkReceivers = new Map<
+    number,
+    NetworkReceiver<unknown>
+  >();
 
   /**
    * An array of network senders
@@ -1462,9 +1437,10 @@ export class FederatedApp extends App {
    * @param networkReceiver The designated network receiver reactor.
    */
   public registerNetworkReceiver(
+    portID: number,
     networkReceiver: NetworkReceiver<unknown>
   ): void {
-    this.networkReceivers.push(networkReceiver);
+    this.networkReceivers.set(portID, networkReceiver);
   }
 
   /**
@@ -1475,7 +1451,7 @@ export class FederatedApp extends App {
   public registerNetworkSender(networkSender: NetworkSender): void {
     this.networkSenders.push(networkSender);
 
-    const portAbsentReaction = networkSender.getLastReactioOrMutation();
+    const portAbsentReaction = networkSender.getLastReactionOrMutation();
     if (portAbsentReaction !== undefined) {
       this.portAbsentReactions.add(portAbsentReaction);
     }
@@ -1699,10 +1675,12 @@ export class FederatedApp extends App {
         // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
         const value: T = JSON.parse(messageBuffer.toString());
 
-        for (const candidate of this.networkReceivers) {
-          if (candidate.getPortID() === destPortID) {
-            candidate.handleMessage(destPortID, value);
-          }
+        try {
+          this.networkReceivers.get(destPortID)?.handleMessage(value);
+        } catch (e) {
+          Log.error(this, () => {
+            return `${e}`;
+          });
         }
       }
     );
@@ -1717,10 +1695,12 @@ export class FederatedApp extends App {
         // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
         const value: T = JSON.parse(messageBuffer.toString());
 
-        for (const candidate of this.networkReceivers) {
-          if (candidate.getPortID() === destPortID) {
-            candidate.handleTimedMessage(destPortID, value, tag);
-          }
+        try {
+          this.networkReceivers.get(destPortID)?.handleTimedMessage(value, tag);
+        } catch (e) {
+          Log.error(this, () => {
+            return `${e}`;
+          });
         }
       }
     );
--- a/src/core/federation.ts
+++ b/src/core/federation.ts
@@ -1209,6 +1209,8 @@ export class FederatedApp extends App {
     }
   }
 
+  //TODO: Add functions for updating MLAA and port statuses
+
   /**
    * Return whether the next event can be handled, or handling the next event
    * has to be postponed to a later time.
@@ -1276,27 +1278,27 @@ export class FederatedApp extends App {
     this.sendRTILogicalTimeComplete(currentTime);
   }
 
-  /**
-   * Enqueue network output control reactions that will send a MSG_TYPE_PORT_ABSENT
-   * message to downstream federates if a given network output port is not present.
-   */
-  protected enqueueNetworkOutputControlReactions(): void {
-    if (
-      this.downstreamFedIDs.length === 0 ||
-      this.outputControlReactionTriggers.length === 0
-    ) {
-      return;
-      // This federate is not connected to any downstream federates via a
-      // logical connection. No need to trigger network output control
-      // reactions.
-    }
-    const trigger = this.outputControlReactionTriggers[0];
-    const event = new TaggedEvent(trigger, this.util.getCurrentTag(), null);
-    Log.debug(this, () => {
-      return "Inserting network output control reaction on reaction queue.";
-    });
-    trigger.update(event);
-  }
+  // /**
+  //  * Enqueue network output control reactions that will send a MSG_TYPE_PORT_ABSENT
+  //  * message to downstream federates if a given network output port is not present.
+  //  */
+  // protected enqueueNetworkOutputControlReactions(): void {
+  //   if (
+  //     this.downstreamFedIDs.length === 0 ||
+  //     this.outputControlReactionTriggers.length === 0
+  //   ) {
+  //     return;
+  //     // This federate is not connected to any downstream federates via a
+  //     // logical connection. No need to trigger network output control
+  //     // reactions.
+  //   }
+  //   const trigger = this.outputControlReactionTriggers[0];
+  //   const event = new TaggedEvent(trigger, this.util.getCurrentTag(), null);
+  //   Log.debug(this, () => {
+  //     return "Inserting network output control reaction on reaction queue.";
+  //   });
+  //   trigger.update(event);
+  // }
 
   protected _finish(): void {
     this.sendRTILogicalTimeComplete(this.util.getCurrentTag());
--- a/src/core/federation.ts
+++ b/src/core/federation.ts
@@ -292,11 +292,12 @@ function isANodeJSCodedError(e: Error): e is NodeJSCodedError {
  * A network sender is a reactor containing a portAbsentReaction.
  */
 export class NetworkSender extends Reactor {
+  
   /**
-   * The last reaction of a NetworkSender reactor is 'portAbsentReactor'.
-   * @returns portAbsentReactor of this NetworkSender reactor
+   * The last reaction of a NetworkSender reactor is the "port absent" reaction.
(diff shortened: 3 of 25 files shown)


src/core/time.ts
```diff
@@ -118,6 +118,9 @@ export class TimeValue {
    * @param other The time value to add to this one.
    */
   add(other: TimeValue): TimeValue {
+    if (other.isNever()) {
+      return this;
+    }
```
hokeun, line 123: This is an interesting definition of adding Never to the TimeValue. However, I think we should be cautious when adding semantics to an operation and see if this definition makes sense. Is this definition used in other targets as well? How about checking if the other time value is Never before applying the add operation and using the time as is rather than defining an addition using never?

lhstrh: Formally, I think `never` is -:infinity:, but here it seems to be treated as zero, which I don't think is quite right. What prompted this change, @byeong-gil?

byeong-gil: In reactor-c, when summing a `tag_t` and an `interval_t`, the `NEVER` interval is treated as 'no delay'. And the 'ZERO' interval adds a microstep.

Here, the `lf_delay_tag` function returns the parameter `tag` as is when the parameter `interval` is `NEVER`.
https://github.com/lf-lang/reactor-c/blob/907ba0e86a3cf7956e96257ab51c2d89c3c71069/core/tag.c#L133-L151

Here, `NEVER` is used to indicate 'no delay'.
https://github.com/lf-lang/lingua-franca/blob/0864dd0287932f994fb3a56e2bc9870d5c63b619/core/src/main/java/org/lflang/federated/extensions/CExtensionUtils.java#L457

Also, I wanted to unify the meaning of these two functions 
https://github.com/lf-lang/reactor-ts/blob/14f5b1352723185a1c8d5834168883ad380f332c/src/core/time.ts#L441-L447
https://github.com/lf-lang/reactor-ts/blob/14f5b1352723185a1c8d5834168883ad380f332c/src/core/time.ts#L120-L130

src/core/federation.ts
```diff
@@ -281,6 +287,125 @@ function isANodeJSCodedError(e: Error): e is NodeJSCodedError {
...
+   */
+  public getLastReactioOrMutation(): Reaction<Variable[]> | undefined {
+    return this._getLastReactionOrMutation();
+  }
+
+  /**
+   * Handle a timed message being received from the RTI.
+   * This function is for NetworkReceiver reactors.
+   * @param portID
+   * @param value
```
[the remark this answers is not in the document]
byeong-gil: This is resolved in the commit [e077972](https://github.com/lf-lang/reactor-ts/pull/186/commits/e0779728b525a32c99b3900a66396d4924720687)

src/core/federation.ts
```diff
@@ -281,6 +287,125 @@ function isANodeJSCodedError(e: Error): e is NodeJSCodedError {
...
+        .asSchedulable(this._getKey(this.networkInputAction))
+        .schedule(0, value);
+    }
+  }
+
+  /**
+   * Handle a timed message being received from the RTI.
+   * This function is for NetworkReceiver reactors.
+   * @param portID
+   * @param value
```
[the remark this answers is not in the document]
byeong-gil: This is resolved in the commit [e077972](https://github.com/lf-lang/reactor-ts/pull/186/commits/e0779728b525a32c99b3900a66396d4924720687)

src/core/federation.ts
```diff
@@ -1376,22 +1454,43 @@ export class FederatedApp extends App {
...
-      federatePortID,
-      federatePortAction
-    );
+    this.networkReceivers.push(networkReceiver);
+  }
+
+  /**
+   * Register a network receiver reactors. It must be registered so it is known by this
+   * FederatedApp and used when add edges for TPO levels.
+   * @param networkSender
```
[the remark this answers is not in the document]
byeong-gil: This is resolved in the commit [e077972](https://github.com/lf-lang/reactor-ts/pull/186/commits/e0779728b525a32c99b3900a66396d4924720687)

src/core/time.ts
hokeun: I think this is ok, but let's wait for @lhstrh 's opinions.

hokeun (approved):
Looks good to me.

src/core/time.ts
byeong-gil: If @lhstrh agrees with this change, How about adding a microstep if the interval is `TimeValue.zero()` to align the function `getLaterTag` with [`lf_delay_tag`](https://github.com/lf-lang/reactor-c/blob/907ba0e86a3cf7956e96257ab51c2d89c3c71069/core/tag.c#L133-L151) in `reactor-c`?
i.e.
```
getLaterTag(delay: TimeValue): Tag { 
  if (this.time.isNever() || delay.isNever()) { 
    return this; 
  } else if (delay.isZero() {
    return new Tag(this.time, this.microstep + 1);
  } else { 
    return new Tag(delay.add(this.time), 0); 
  } 
} 
```

lhstrh: Hmm.. I'm frankly a bit puzzled by the examples in the C runtime/code generator that @byeong-gil is sharing. Treating `NEVER` as "no delay", for instance, doesn't sound right to me. Perhaps @edwardalee can shed some light on this.

__tests__/connection.test.ts
```diff
@@ -39,12 +39,6 @@ describe("Check canConnect", () => {
           true
         );
       });
-
-      it("canConnect failure", () => {
-        expect(this.canConnect(this.destination.in, this.source.out)).toBe(
```
lhstrh, line 44: Leaving a marker for @axmmisaka to review.

__tests__/hierarchy.ts
```diff
@@ -40,32 +40,20 @@ describe("Container to Contained", () => {
     expect(
       app.container.canConnect(app.container.contained.a, app.container.a)
     ).toBe(false);
-    expect(
-      app.container.canConnect(
-        app.container.contained.a,
```
lhstrh, line 45: Leaving a marker for @axmmisaka to review.

src/core/federation.ts
```diff
@@ -281,6 +287,125 @@ function isANodeJSCodedError(e: Error): e is NodeJSCodedError {
   return typeof (e as NodeJSCodedError).code === "string";
 }
 
+/**
+ * A network reactor is a reactor handling network actions (NetworkReceiver and NetworkSender).
```
lhstrh, line 291: Is there any reason to have `NetworkSender` and `NetworkReceiver` have a common base class? It seems to me they should have very little in common.

lhstrh: A more concrete question would be: why not define a `NetworkSender` and `NetworkReceiver` class instead of this common `NetworkReactor` class?

byeong-gil: They should have a common parent class so that we can use them when analyzing dependencies implied by TPO levels. **However**, TPO-level things are not included in this branch, so I'll define a `NetworkSender` and `NetworkReceiver` class instead of the common class.

src/core/time.ts
edwardalee: The reason we treat NEVER as no delay is that we have to make a distinction between zero delay and no delay. interval_t is an int64, and there is only one encoding for zero.  I think the reasoning was that an interval of minus infinity doesn't seem very useful. It is the time between a point in time and the time something never happened (?) ... a bit of a brain twister... Like trying to figure out when the big bang happened (hint: it happened NEVER, as you can determine by working time backwards and applying relativity as gravitation increases).

src/core/federation.ts
byeong-gil: Now, network receiver and sender reactors are instantiated as `NetworkReceiver` and `NetworkSender`, respectively.  For more details, please refer to the commit [e36fb0b](https://github.com/lf-lang/reactor-ts/pull/186/commits/e36fb0b4658e1f469ec37dfcced754e291b9982d).

src/core/federation.ts
```diff
@@ -281,6 +287,123 @@ function isANodeJSCodedError(e: Error): e is NodeJSCodedError {
...
 
+/**
+ * A network sender is a reactor containing a portAbsentReaction.
+ */
+export class NetworkSender extends Reactor {
+  /**
+   * The last reaction of a NetworkSender reactor is 'portAbsentReactor'.
+   * @returns portAbsentReactor of this NetworkSender reactor
+   */
+  public getLastReactioOrMutation(): Reaction<Variable[]> | undefined {
```
lhstrh, line 298: The name of this method has a typo. But aside from that, why is it needed? This looks like a "no-op" override...

src/core/federation.ts
```diff
@@ -281,6 +287,123 @@ function isANodeJSCodedError(e: Error): e is NodeJSCodedError {
...
+}
+
+/**
+ * A network receiver is a reactor handling a network input.
+ */
+export class NetworkReceiver<T> extends Reactor {
+  /*
+   * A FederatePortAction instance of this NetworkReceiver.
+   */
+  private networkInputAction: FederatePortAction<T> | undefined = undefined;
```
lhstrh, line 310: I don't think you need to set it to `undefined` explicitly?
```suggestion
  private networkInputAction: FederatePortAction<T> | undefined;
```

src/core/federation.ts
```diff
@@ -281,6 +287,123 @@ function isANodeJSCodedError(e: Error): e is NodeJSCodedError {
...
+    // This message is untimed, so schedule it immediately.
+    if (portID !== this.portID) {
+      this.util.reportError(
+        "FederatedApp attempts to pass the tagged message to the wrong port ID"
+      );
+      return;
+    }
+    if (this.networkInputAction !== undefined) {
+      this.networkInputAction
+        .asSchedulable(this._getKey(this.networkInputAction))
```
lhstrh, line 353: Since `networkInputAction` is `private`, why don't we store the schedulable action instead of retrieving it every time we schedule?

src/core/federation.ts
```diff
@@ -1597,63 +1691,36 @@ export class FederatedApp extends App {
...
           return "(Untimed) Message received from RTI.";
         });
         // TODO (axmmisaka): use a type-safe interface, like io.ts?
         // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
         const value: T = JSON.parse(messageBuffer.toString());
 
-        destPortAction
-          .asSchedulable(this._getKey(destPortAction))
-          .schedule(0, value);
+        for (const candidate of this.networkReceivers) {
```
lhstrh, line 1702: This is OK if we expect the number of receivers to always be small, but a map would be more efficient.

src/core/reactor.ts
```diff
@@ -1202,7 +1202,8 @@ export abstract class Reactor extends Component {
         }
       } else {
         // IN to OUT
-        return false;
+        // FIXME: Make this direct feedthrough false after the generated code deletes it.
```
lhstrh, line 1205: I don't quite understand this comment.

src/core/reactor.ts
```diff
@@ -2141,7 +2142,7 @@ export class App extends Reactor {
   /**
    * Priority set that keeps track of reactions at the current Logical time.
    */
-  private readonly _reactionQ = new ReactionQueue();
+  protected readonly _reactionQ = new ReactionQueue();
```
lhstrh, line 2145: Are you sure this is necessary? If so, why?

__tests__/connection.test.ts
axmmisaka: I don't recommend deleting it, but instead setting this to expecting `true` instead. 
What I am more worried about is that this is directly related to nothing but https://github.com/lf-lang/reactor-ts/pull/186/files#r1275393168, where we change from always disallowing `in` to `out` connection to always allowing.
If there's a FIXME, I recommend documenting what's wrong in an issue, and write relevant tests in a new PR that shows the expected behaviours (which might be failing, but it's okay, because we need to address the FIXME).
But just don't remove these tests because they are failing.

```mermaid
    graph
     0[/"testApp[M0]"/]
     subgraph "testApp.fc5128bb"
      1[/"testApp.fc5128bb.OutPort"\]
      2[/"testApp.fc5128bb[M0]"/]
     end
     subgraph "testApp.9d6363af"
      3[\"testApp.9d6363af.InPort"/]
      4[\"testApp.9d6363af.InPort"/]
      5[/"testApp.9d6363af[M0]"/]
     end
    
    0 --> 2
    0 --> 5
    3 --x|Allow this????| 1
```

__tests__/hierarchy.ts
axmmisaka: Ibidem - et seq.

src/core/federation.ts
axmmisaka: Because `_getLastReactionOrMutation` in `Reactor` is `protected` but for `NetworkSender` it needs to be exposed.

https://github.com/lf-lang/reactor-ts/blob/08fa08468d2c45714cd8e1fe6bd24f83edadda89/src/core/federation.ts#L1475-L1482

src/core/federation.ts
axmmisaka: It depends, but it either needs to be initialised here, or in `constructor`. I suggest moving it into constructor, because `portID` is initialised there as well.

src/core/reactor.ts
axmmisaka: It's used by `federatedApp`, a subclass of `Reactor`

 https://github.com/lf-lang/reactor-ts/blob/08fa08468d2c45714cd8e1fe6bd24f83edadda89/src/core/federation.ts#L1488-L1492

src/core/time.ts
lhstrh: > If @lhstrh agrees with this change, How about adding a microstep if the interval is `TimeValue.zero()` to align the function `getLaterTag` with [`lf_delay_tag`](https://github.com/lf-lang/reactor-c/blob/907ba0e86a3cf7956e96257ab51c2d89c3c71069/core/tag.c#L133-L151) in `reactor-c`? i.e.
> 
> ```
> getLaterTag(delay: TimeValue): Tag { 
>   if (this.time.isNever() || delay.isNever()) { 
>     return this; 
>   } else if (delay.isZero() {
>     return new Tag(this.time, this.microstep + 1);
>   } else { 
>     return new Tag(delay.add(this.time), 0); 
>   } 
> } 
> ```

This looks OK to me, except I think that we should handle the "forever" case specially as well, because if we don't, that would lead to overflow. Concretely, I think that we should have :infinity: + something = :infinity:.

axmmisaka: One thing that confused me was that `NEVER` was defined as `0x8000000000000000` in reactor-c (that is something like negative infinity in `int64`???) - I guess it's just a practical choice. I need to bang me head several times on the was again to understand this (probably), but is it correct to conclude that practically no delay introduces no microstep change whereas zero delay does?

axmmisaka (commented):
Other than on the tests, I only have comments.

src/core/time.ts
lhstrh: > I need to bang me head several times on the was again to understand this (probably), but is it correct to conclude that practically no delay introduces no microstep change whereas zero delay does?

This is also what I'm unsure about, actually. If we treat -:infinity: as zero, then adding it to some other tag should yield the same result as adding zero to it. This suggests that we should have
```
getLaterTag(delay: TimeValue): Tag { 
  if (this.time.isNever() || delay.isNever() || delay.isZero()) { 
    return new Tag(this.time, this.microstep + 1);
  if (this.time.isForever() || delay.isForever()) {
    return forever();
  } else { 
    return new Tag(delay.add(this.time), 0); 
  }
} 
```

WDYT, @edwardalee?

src/core/time.ts
```diff
@@ -118,6 +118,9 @@ export class TimeValue {
    * @param other The time value to add to this one.
    */
   add(other: TimeValue): TimeValue {
+    if (other.isNever()) {
```
lhstrh, line 121: If we agree that `t` + -:infinity: = `t`, then -:infinity: + `t` = `t` should also hold. This latter case is not handled currently.

As stated below, I think we should also ensure that `t` + :infinity: = :infinity: + `t` = :infinity:.

src/core/time.ts
byeong-gil: I have a concern about the comment
>then adding it to some other tag should yield the same result as adding zero to it.

Because, @edwardalee said that
>we have to make a distinction between zero delay and no delay.

More concretely, when a federate sends two messages at the tag (0, 0) with `NEVER` and `ZERO` additional delays, the destination federate schedules them at the tag (0, 0) and (0, 1) respectively. Don't we have to distinguish those no and zero delays?

src/core/federation.ts
byeong-gil: As @axmmisaka said, I set it to `undefined` to initialize it. I'll move it into constructor.

src/core/reactor.ts
byeong-gil: @axmmisaka is right. Let me supplement it a little bit. 
From now on, `portAbsentReaction`s don't have a trigger. So we have to push those reactions to `_reactionQ` directly in the `FederateApp`. Also, I think this is safe because `_reactionQ` is owned by `App` and only `FederatedApp` is a subclass of `App`.

__tests__/connection.test.ts
byeong-gil: Thanks to @axmmisaka, I found that 
https://github.com/lf-lang/reactor-ts/blob/14f5b1352723185a1c8d5834168883ad380f332c/src/core/reactor.ts#L1205
this line cannot be simply turned into `true`. 

In the commit [01624c4](https://github.com/lf-lang/reactor-ts/pull/186/commits/01624c4a842451937a4e265ef8f7d61eb9944ee9), the connection between input and output ports are allowed only if those are in the same container and the container calls `canConnect` to connect them.
https://github.com/lf-lang/reactor-ts/blob/14965ac762988c2608913feaa8b6e262fc7dddab/src/core/reactor.ts#L1204-L1213
Also, I added a test related to this.
https://github.com/lf-lang/reactor-ts/blob/14965ac762988c2608913feaa8b6e262fc7dddab/__tests__/hierarchy.ts#L43-L48

src/core/federation.ts
byeong-gil: Yes, `NetworkSedner` does not extend `FederatedApp`, and `FederatedApp` has to access the `NetworkSender`'s last reaction because it's a port absent reaction.

src/core/federation.ts
byeong-gil: A map is now used to store `networkReceiver`s in the commit [b8b6c73](https://github.com/lf-lang/reactor-ts/pull/186/commits/b8b6c73d48ea9c28b2ebcd0bd9fd1adaad84a891).

src/core/reactor.ts
lhstrh: OK, I agree that it's OK to make the visibility `protected`,

But if these reactions do not have a trigger, then how are they assigned a priority? Or are we adding these to the reaction queue only once we're ready to end of a tag, when all inputs are known and the reaction queue is empty?

src/core/reactor.ts
byeong-gil: I made some changes and left the description of them in this comment https://github.com/lf-lang/reactor-ts/pull/186#discussion_r1275695631.

src/core/reactor.ts
byeong-gil: AFAIK, priorities are assigned to reactions, not triggers, right? Assuming that this is right, I think reactions are sorted by their priorities when they are pushed to `_reactionQ`. Currently, `portAbsentReaction`s are pushed into `_reactionQ` before executing reactions at each tag.

https://github.com/lf-lang/reactor-ts/blob/825f8fe3ea973c185be40de458583d3dcb00862b/src/core/reactor.ts#L2393-L2397

src/core/time.ts
lhstrh: I think I'm getting it now, but I think we're conflating two issues.

1. How do we determine the tags of events sent from one federate to another (this depends on the delay along the connection). 
2. How does the arithmetic of tags work.

As for 1, I see no reason why we need to encode "no delay" as `NEVER`, as is done in the C implementation; in TypeScript, we can just record it as `undefined` or `null`.

As for 2, my view is expressed in https://github.com/lf-lang/reactor-ts/pull/186/files#r1275558491

src/core/reactor.ts
lhstrh: It had escaped me that the reactions in the `NetworkSender` are code-generated. I'm assuming it works because the "port absent" reaction depends on the "port present" reaction that is triggered by the input? But can't they share the same trigger then? I haven't yet looked at the code generator carefully...

__tests__/hierarchy.ts
byeong-gil: I made some changes related to this and left the description of them in this comment https://github.com/lf-lang/reactor-ts/pull/186#discussion_r1275695631.

src/core/reactor.ts
byeong-gil: >the "port absent" reaction depends on the "port present" reaction that is triggered by the input?

That's right. There is an edge that represents it in the `dependencyGraph`.
>But can't they share the same trigger then? 

For this question, I would share @petervdonovan's answer to the same question of mine. 

>That "port absent" reaction has to be triggered when msg is not present because its purpose is to send a message to notify the receiving federate that the port is absent in the current tag. So it would not work to have it be triggered by msg.

src/core/federation.ts
lhstrh: OK, I just renamed the method to reflect that it's just meant to retrieve the "port absent" reaction.

src/core/time.ts
lhstrh: Note that if we don't encode "nothing" as -:infinity:, then I think we should have `t` + -:infinity: = -:infinity: + `t` = -:infinity: also.

src/core/federation.ts
byeong-gil: This is solved in the commit [825f8fe](https://github.com/lf-lang/reactor-ts/pull/186/commits/825f8fe3ea973c185be40de458583d3dcb00862b)

lhstrh: Addressed in https://github.com/lf-lang/reactor-ts/pull/186/commits/825f8fe3ea973c185be40de458583d3dcb00862b.

lhstrh: You beat me to it :rofl:

byeong-gil: I won! :laughing:

src/core/time.ts
byeong-gil: This is addressed in [26d0a9f](https://github.com/lf-lang/reactor-ts/pull/186/commits/26d0a9f240c3619d9d8b899d163697047116bd7c). `t` + -:infinity: = -:infinity: + `t` = -:infinity:, `t` + ♾️ = ♾️ + `t` = ♾️, and -:infinity: + ♾️ = 0 now. Additionally, 'no delay' is represented as `undefined` and overflow is handled either.

src/core/time.ts
byeong-gil: 'no delay' is expressed by `undefined` in [26d0a9f](https://github.com/lf-lang/reactor-ts/pull/186/commits/26d0a9f240c3619d9d8b899d163697047116bd7c). Also, `NEVER` and `FOREVER` is treated as -♾️ and ♾️, unlike `reactor-c`. 
I left a FIXME in case we will add a microstep when we add `ZERO` to a tag `t` like `reactor-c`. 
https://github.com/lf-lang/reactor-ts/blob/2b36efb609c67f29bee1fedfbefa2b280424f253/src/core/time.ts#L470-L477

src/core/time.ts
```diff
@@ -118,6 +118,19 @@ export class TimeValue {
    * @param other The time value to add to this one.
    */
   add(other: TimeValue): TimeValue {
+    if (
+      (this.isNever() && other.isForever()) ||
```
lhstrh, line 122: We don't have to treat this case specially, I think.
We have
```
static never(): TimeValue {
    return new TimeValue(Number.MIN_SAFE_INTEGER, 0);
  }

  static forever(): TimeValue {
    return new TimeValue(Number.MAX_SAFE_INTEGER, 0);
  }
```
And `Number.MAX_SAFE_INTEGER` + `Number.MIN_SAFE_INTEGER` = `0`.

lhstrh: Never mind this comment. What you did is correct. I've added some unit tests to verify.

src/core/time.ts
```diff
@@ -126,7 +139,14 @@ export class TimeValue {
       seconds += 1;
       nanoseconds -= TimeUnit.sec;
     }
-    return TimeValue.secsAndNs(seconds, nanoseconds);
+
+    // Check an overflow.
+    if (!Number.isSafeInteger(seconds)) {
+      // The overflow happens.
+      return TimeValue.forever();
```
lhstrh, line 146: We currently don't allow negative time intervals. If we decide to allow them at some point, we have to pay attention to the direction of the overflow, but for now, this works.

src/core/time.ts
```diff
@@ -202,6 +222,14 @@ export class TimeValue {
     }
   }
 
+  isForever(): boolean {
+    if (this.seconds === Number.MAX_SAFE_INTEGER && this.nanoseconds === 0) {
```
lhstrh, line 226: I don't think we care about the nanoseconds. If we say that `(Number.MAX_SAFE_INTEGER, 0)` is :infinity: then shouldn't `(Number.MAX_SAFE_INTEGER, 0)` also be :infinity:?
```suggestion
    if (this.seconds === Number.MAX_SAFE_INTEGER) {
```

src/core/time.ts
```diff
@@ -435,11 +463,13 @@ export class Tag {
...
-   * unchanged with its current `microstep`.
+   * unchanged with its current `microstep`. If the delay is undefined,
+   * the tag is returned as it was.
    * @param delay The time interval to add to this time instant.
    */
-  getLaterTag(delay: TimeValue): Tag {
-    if (delay.isZero() || delay.isNever()) {
+  getLaterTag(delay: TimeValue | undefined): Tag {
+    if (delay === undefined || delay.isZero()) {
+      // FIXME: Should a tag + zero returns one microstep after the tag?
```
lhstrh, line 472: Yes, here we definitely should add a microstep. The purpose of this method is to supply the runtime with a tag that is _later_ and thus needs to be strictly greater than the current one.

lhstrh: Changing this method to add a microstep yields a lot of failures. My best guess is that we're adding a microstep elsewhere already...

lhstrh: > Changing this method to add a microstep yields a lot of failures. My best guess is that we're adding a microstep elsewhere already...

Yes, we do: https://github.com/lf-lang/reactor-ts/blob/e8e6a4021b8a5a59decce758ccbb882706cdb6fd/src/core/action.ts#L136

lhstrh (approved):

src/core/time.ts
lhstrh: Marking this as resolved.

PR #198, opened 2023-07-24
Handle cyclic dependencies by applying the `MLAA` based execution and `TPO level`

byeong-gil:
Relevant LF PR: https://github.com/lf-lang/lingua-franca/pull/1925

This PR tries to handle cyclic dependencies by adopting TPO level and MLAA concepts (a more concrete description of those concepts is here: https://github.com/lf-lang/lingua-franca/pull/1684).

To handle cyclic connections properly, the below tasks are required.

- [x] Add edges for TPO levels so that the dependency graph recognizes dependencies between network reactors.
- [x] Update MLAA according to information from port-absent messages and PTAG. 
- [x] Prevent executing reactions with priorities larger or equal to MLAA.

In addition, this PR also addresses the below problems.
* The `ReactionGraph` is now containing every reaction node of the dependency graph.
* The function `getLaterTag` adds a microstep when adding a `ZERO` to a tag.
* Physical actions are scheduled by the current physical time.


byeong-gil:
[`SimpleFederated.lf`](https://github.com/lf-lang/lingua-franca/blob/ts-cyclic-dependencies/test/TypeScript/src/federated/SimpleFederated.lf) is failing due to the 'missing reactions situation' caused by the constructor of  [SortablePrecedenceGraph](https://github.com/lf-lang/reactor-ts/blob/8ef0e578d959af68855058550ffe9d6a11566ecf/src/core/graph.ts#L357-L370).
If I understand correctly, `ReactionGraph` that extends `SortablePrecedenceGraph` is a graph containing only the `reaction` nodes from `dependencyGraph`. 
https://github.com/lf-lang/reactor-ts/blob/8ef0e578d959af68855058550ffe9d6a11566ecf/src/core/graph.ts#L389-L396
However, in the constructor above, the `reaction` nodes can be missed. For example, If there is a dependency like `reaction1 -> reaction2 -> port1`, `port1` is a leaf but not has a type of `reaction`. Thus, `reaction1` and `reaction2` are failed to enter the `ReactionGraph`.

src/core/action.ts
```diff
@@ -79,20 +79,11 @@ export class Action<T> extends ScheduledTrigger<T> implements Read<T> {
...
-        // on the platform, then the timestamp becomes the current physical time.
-        // Otherwise the tag is computed like a logical action's tag.
-
-        const physicalTime = getCurrentPhysicalTime();
-        if (tag.time.isEarlierThan(physicalTime)) {
-          tag = new Tag(getCurrentPhysicalTime(), 0);
-        } else {
-          tag = tag.getMicroStepsLater(1);
-        }
+        tag = new Tag(getCurrentPhysicalTime(), 0).getLaterTag(delay);
```
byeong-gil, line 82: I changed physical actions to be scheduled with the current physical time. Let me explain why this is more reasonable. 
First, the delay is added to the logical tag instead of the physical time. The test [DistributedCountPhysicalAfterDelay.lf](https://github.com/lf-lang/lingua-franca/blob/ts-cyclic-dependencies/test/TypeScript/src/federated/DistributedCountPhysicalAfterDelay.lf) is failed because of that. In the test, the reactor `Count` sends a message via the physical connection at the logical tag (200 msec, 0). The physical connection has a delay of `400 msec`. The test specifies that the destination should schedule the message at a time larger than `600 msec`. Nonetheless, in the previous operation, it was scheduled at the tag (400 msec, 1). Because the 'current logical time (0 sec) + delay (400 msec)' = `400 msec` is larger than the current physical time.
Second, `reactor-c` also schedules physical actions like this. ([`lf_schedule`](https://github.com/lf-lang/reactor-c/blob/ccf230f3d1027ee39987066b9e1f3211560e9f6e/core/reactor_common.c#L939-L941)).

@lhstrh Do you think this is ok?

edwardalee:
I think that an after delay on a physical connection should be added to the physical time, not the logical time, which is what this test seems to be checking. In any case, the behavior is not documented here https://www.lf-lang.org/docs/handbook/composing-reactors?target=lf-c#physical-connections
But it should be.

byeong-gil:
> I think that an after delay on a physical connection should be added to the physical time, not the logical time, which is what this test seems to be checking. In any case, the behavior is not documented here https://www.lf-lang.org/docs/handbook/composing-reactors?target=lf-c#physical-connections But it should be.

Thank you for answering on my question, @edwardalee! `reactor-ts` is now adding an after delay on a physical action to the physical time.

src/core/action.ts
byeong-gil: Mark this as resolved as @edwardalee confirmed the concept of this change.

lhstrh:
> [`SimpleFederated.lf`](https://github.com/lf-lang/lingua-franca/blob/ts-cyclic-dependencies/test/TypeScript/src/federated/SimpleFederated.lf) is failing due to the 'missing reactions situation' caused by the constructor of [SortablePrecedenceGraph](https://github.com/lf-lang/reactor-ts/blob/8ef0e578d959af68855058550ffe9d6a11566ecf/src/core/graph.ts#L357-L370). If I understand correctly, `ReactionGraph` that extends `SortablePrecedenceGraph` is a graph containing only the `reaction` nodes from `dependencyGraph`.
> 
> https://github.com/lf-lang/reactor-ts/blob/8ef0e578d959af68855058550ffe9d6a11566ecf/src/core/graph.ts#L389-L396
> 
> However, in the constructor above, the `reaction` nodes can be missed. For example, If there is a dependency like `reaction1 -> reaction2 -> port1`, `port1` is a leaf but not has a type of `reaction`. Thus, `reaction1` and `reaction2` are failed to enter the `ReactionGraph`.

It's the "collapsed" graph that only has reactions in it. The precedence graph that is obtained from the main reactor (before it is collapsed) is of type `PrecedenceGraph<Port<unknown> | Reaction<Variable[]>>` and thus contains both reactions _and_ ports.

src/core/action.ts
lhstrh: Hm, I'm not sure that the behavior you implemented is the behavior we want, even though it appears to match what happens in the C implementation. I'm actually having a hard time coming up with a use case for adding a logical time delay to a physical connection. @edwardalee, do we have one? What's currently implemented in `master` of `reactor-ts` is more like a "best effort" logical connection, which I can imagine being useful.

edwardalee: I'm not sure about this.  When lf_schedule gets called, what is the meaning of "current logical time"?  I think the right approach here is to provide documentation, I.e., a specification.  We need a clear definition, and then we need to make sure the tests check that the definition is what we have implemented.

So, can you explain exactly what an after delay on a physical connection means?

byeong-gil: I used 'an after delay on a physical connection' to explain the delay of the expression `a.out ~> b.in after delay`.

lhstrh: > I'm not sure about this. When lf_schedule gets called, what is the meaning of "current logical time"? I think the right approach here is to provide documentation, I.e., a specification. We need a clear definition, and then we need to make sure the tests check that the definition is what we have implemented.

I think that our stake in the ground is that federation should be semantics preserving. 

In an unfederated context, the tag would be based on the current physical time, but when the event travels from one federate to the other, there are two physical times to consider: the one at the time of sending, `Ts`, and the one at the time of receiving, `Tr`. It seems natural to add the delay to the time of _sending_, but then there is a corner case to handle: where the delay incurred by transport + clock skew is greater than the delay on the connection, which would result in the time being less the current physical time at the receiver, which in which case the receiver might already be ahead.

So perhaps the tag at the receiver should be `(max(Ts + d, Tr), 0)`? It could also be `(Tr+d, 0)`, as @byeong-gil implemented, but it seems strange to add a delay to an event originating from a physical action but then ignore the physical time that elapses while the event is in transit. Note that the `(max(Ts + d, Tr), 0)` case also ensure that the tag of the event always has a time value greater or equal to `Ts`. In the `(Tr+d, 0)` case, when `Tr < Ts` due to clock skew, we could get a reaction to an event at a time earlier than when the event was produced (if the clock skew is greater than the network latency). This, too, seems strange, and is probably the strongest reason I can come up with to not opt for that solution.

That being said, do we have any practical use for physical connections with delays? Even in unfederated programs?

byeong-gil: I have questions about `((max(Ts + d, Tr), 0)`. For explanation, let's say that `tc` is a clock skew and `tn` is a network latency. Also, `Tr` is the physical time of the receiver at the time the sender sends the message. Thus, `Ts = Tr + tc`. In this case, we have to select between `((max(Ts + d, Tr + tn), 0)` and `(Tr + tn + d, 0)`.

Screenshot_20230806_134741_Samsung Notes

**First**, I couldn't imagine the clear specification of the delay `d`. When we use `(Tr + tn + d, 0)`, the 'actual delay' (the delay between the message sending and the action scheduling) is apparently `tn + d`, which is 'network delay + additional delay'. [HelloWorldPhysicalAfter.lf#L7-L8](https://github.com/lf-lang/playground-lingua-franca/blob/3af88a78c2576834c1204ae2c25695f0c9088ce7/examples/C/src/distributed/HelloWorldPhysicalAfter.lf#L7-L8) also describes like that. So we can use `d` when we need additional delay among the network delay. However, when we use `((max(Ts + d, Tr + tn), 0)`, the 'actual delay' becomes `max((tc + d), tn)`. What would be the definition of `d`? 


**Second**, in the https://www.lf-lang.org/docs/handbook/composing-reactors?target=lf-c#physical-connections, 
>Physical connections are useful sometimes in [Distributed-Execution](https://www.lf-lang.org/docs/handbook/distributed-execution) in situations where the nondeterministic logical delay is tolerable.

And I think that the corner case you described below can be included in `tolerable nondeterministic logical delay`, doesn't it? Because, in my view, normally the network latency is greater than the clock skew. Please fix me if I'm wrong.
>In the (Tr+d, 0) case, when Tr < Ts due to clock skew, we could get a reaction to an event at a time earlier than when the event was produced (if the clock skew is greater than the network latency).


**Lastly**, How can the receiver know the `Ts`? I'm curious about this because in https://www.lf-lang.org/docs/handbook/composing-reactors?target=lf-c#physical-connections, the message, 
>Such connections are more efficient because timestamps need not be transmitted

lhstrh: To clarify, I meant for `Tr` to denote the physical time at which the event is _received_. There is no mechanism to find out what the time at the receiver is at the moment that it is sent something by another federate. Also, note that `d` and `tn` are always positive, but `tc` can be negative. My point was that the arrow you drew in the sequence diagram (where I assume the y-axis denotes physical time as recorded at each federate), could point upwards if `tn + tc < d` (assuming `Tr = Ts + tn + tc`), meaning that the event is seen by the receiver at a tag earlier than the tag at which it was produced by the sender. I did not write the documentation that you cite, but it confirms this observation.

In order to implement `(max(Ts + d, Tr), 0)`, you need to know `Ts` (`d` is just the after delay), meaning it would have to be sent along with the message, and I agree with you that this might defeat the purpose of physical connections. I don't question the utility of physical connections, but I am unsure of the utility of _delays_ along physical connections. If there is tolerance for nondeterminism in the first place, then why use a logical delay at all?

Certainly, if we don't have good use cases for it, I won't advocate implementing something complicated like `(max(Ts + d, Tr), 0)` and would stick with the simple `(Tr+d, 0)` solution. @edwardalee, are you aware of any use cases for delayed physical connections?

edwardalee: I have no use case and would stick with the simple solution.

byeong-gil:
I marked this PR as a "draft" again because I found non-determinism errors in this PR.

src/core/federation.ts
```diff
@@ -1347,6 +1403,44 @@ export class FederatedApp extends App {
     return true;
   }
 
+  /**
+   * @override
+   * Iterate over all reactions in the reaction queue and execute them.
+   * If a reaction's priority is less than MLAA, stop executing and return.
+   * @returns Whether every reactions at this tag are executed.
+   */
+  protected _react(): boolean {
```
byeong-gil, line 1412: @lhstrh I removed the call of `_updateMaxLevel` here and the function is called only when it is necessary (please refer to the commit [a413a9f](https://github.com/lf-lang/reactor-ts/pull/198/commits/a413a9f9da76572e0a570b06358e11ee0274204a)).

One question is, if I remember correctly, you also wanted to divide this `_react` function, right? Nonetheless, I don't know how to separate this function. Because this function has a simple role of doing react until the reaction of `_reactionQ` is allowed to execute. 

A possible separation is, pushing executable reactions into a new list (in the new function) and then processing them (in the `_react()`). However, I and @hokeun think that the current state is also clear. What do you think about?

src/core/action.ts
```diff
@@ -79,20 +79,11 @@ export class Action<T> extends ScheduledTrigger<T> implements Read<T> {
...
       } else if (this.action instanceof FederatePortAction) {
         if (intendedTag === undefined) {
           throw new Error(
-            "FederatedPortAction must have an intended tag from RTI."
+            "Logical FederatedPortAction must have an intended tag from RTI."
```
lhstrh, line 86:
```suggestion
            "No intended tag given while attempting to schedule an event coming from another federate."
```

src/core/federation.ts
lhstrh: I agree that this is fine. I think that maybe I didn't realize that this was overriding the `_react` function of `App`. That said, it looks like there are currently some federation-related functions in `App` that should probably be moved into `FederatedApp`. But this can be done in a separate PR...

src/core/graph.ts
```diff
@@ -392,6 +392,11 @@ export class SortablePrecedenceGraph<
         this.addNode(leaf);
         search(leaf, pg.getUpstreamNeighbors(leaf));
         visited.clear();
+      } else {
+        // leaf is not a type. It's upstream neighbors become a new leafs.
+        for (const newLeaf of pg.getUpstreamNeighbors(leaf)) {
+          leafs.add(newLeaf);
```
lhstrh, line 398: This adds entries to a set already being iterated over, and I don't think the iterator gets updated, so I suspect there are corner cases in which this doesn't work. Shouldn't the call to `search` simply be outside of the `if` block? Also, should we really be clearing the `visited` set at all?

lhstrh (commented):
Great work! I left some feedback...

src/core/graph.ts
byeong-gil: The intended behavior of this part is like below.
```mermaid
graph
0[port0]
1[reaction0]
2[reaction1]
3[port1]
4[other upstreams...]
5[upstreams...]

1 --> 0
3 --> 0
2 --> 3
4 -->|"search(reaction0, other upstreams...)"| 2
5 -->|"search(reaction0, upstreams...)"| 1
```
Before this modification, `reaction0` and `reaction1` are just ignored as `port0` is `leaf` and is not an instance of `reaction`.

And I agree with
>This adds entries to a set already being iterated over, and I don't think the iterator gets updated, so I suspect there are corner cases in which this doesn't work.

Particularly, in case multiple leaves share the same upstream node. For instance, if we have a dependency graph like below, `port0` is added twice.
```mermaid
graph
0[port0]
1[reaction2]
2[reaction3]
3[port1]
4[other upstreams...]
5[upstreams...]
6[reaction0]
7[reaction1]

1 --> 0
3 --> 0
2 --> 3
4 --> 2
5 --> 1
0 --> 6
0 --> 7
```
Nevertheless, I don't think 
>Shouldn't the call to search simply be outside of the if block? 

will work. Because the first parameter of `search` has to have a type of `T` (`T` is the same as `type`). 
So, how about adding upstreams of the leaf to `leafs` if those are not visited yet?

Also, about the below comment,
>Also, should we really be clearing the visited set at all?

I'm with you that that part is not necessary. However, I'm not 100% sure as I didn't write that code. @axmmisaka Could you please give your opinion about deleting this line? https://github.com/lf-lang/reactor-ts/blob/05480e333401b16149b1b7e7da73bf8bdf6d86ba/src/core/graph.ts#L394

lhstrh: I think I may have written that code :rofl:

I pushed some changes to clarify what the code does...

src/core/federation.ts
```diff
@@ -1712,33 +2035,62 @@ export class FederatedApp extends App {
...
-        /*
-                this.greatestTimeAdvanceGrant = tag;
-                this._isLastTAGProvisional = true;
-                this._requestImmediateInvocationOfNext();
-                */
-        //
+      // Update the greatest time advance grant and update MLAA.
+      this.greatestTimeAdvanceGrant = ptag;
+      this._isLastTAGProvisional = true;
+      if (!this._active) {
```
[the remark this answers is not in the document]
lhstrh: :+1:

lhstrh (approved):

src/core/graph.ts
```diff
@@ -373,25 +373,32 @@ export class SortablePrecedenceGraph<
...
-        visited.clear();
+
+    for (const node of startNodes) {
+      if (node instanceof type) {
+        this.addNode(node);
+        search(node, pg.getDownstreamNeighbors(node));
+      } else {
+        // Look further upstream for start nodes that match the type.
+        for (const newStartNode of pg.getDownstreamNeighbors(node)) {
+          startNodes.add(newStartNode);
```
byeong-gil, line 400:
```suggestion
          if (!visited.has(newStartNode)) {
            startNodes.add(newStartNode);
          }
```
@lhstrh How about checking whether the further upstream is visited or not for efficiency?

PR #210, opened 2023-08-10
Use dependabot grouping feature to make sure it doesn't make 1000 PRs every day

axmmisaka:
It is a beta feature in public beta released a month ago, which is something we desperately need.
https://github.blog/changelog/2023-06-30-grouped-version-updates-for-dependabot-public-beta/

Demo: https://github.com/axmmisaka/reactor-ts/pull/21


lhstrh (approved):
This is useful indeed. Thanks!

Issue #112, opened 2022-07-24
Adjusting access to utils for federated execution


hokeun:
After fed-gen is done, the generated .lf files from a federated LF program will be the main reactors, which is currently generated as `App` class rather than `FederatedApp` class.
However, inside the network sender body (reaction body), `this.util.sendRTITimedMessage` will be called in the main reactor. This is currently not possible since `App` class doesn't implement sendRTIMessage or sendRTITimedMessage like this.

```
    protected sendRTIMessage<T extends Present>(data: T, destFederateID: number, destPortID: number) {
        throw new Error("Cannot call sendRTIMessage from an App. sendRTIMessage may be called only from a FederatedApp");
    }

    protected sendRTITimedMessage<T extends Present>(data: T, destFederateID: number, destPortID: number) {
        throw new Error("Cannot call sendRTIMessage from an App. sendRTIMessage may be called only from a FederatedApp");
    }
```

Possible solutions:

1. Somehow we distinguish a reactor to be generated as `FederatedApp` and a reactor to be generated as `App`.
2. Allow sendRTITimedMessage to be called inside `App`
3. Other suggestions? (@lhstrh)

ByeongGil-Jun:
@marten Below is what I found about waiting `networkInputControlReaction` by using `async`/`await`. It would be a great pleasure if you read it and tell me your thoughts.

There are two problems. 
First, how can we suspend and control back in the right way? 
Second, how can we scan and process the RTI messages while the `_react` is suspended?

For the first question, we should make all upper functions of `_react()` be an async function and put `await` carefully.
These are four scenarios that `_react()` is called.
1. [`_startExecuting`](https://github.com/lf-lang/reactor-ts/blob/554a86128c35c76335dec50d2c953590fb02d259/src/core/reactor.ts#L2339) → `_react()`
2. [`_startExecuting`](https://github.com/lf-lang/reactor-ts/blob/554a86128c35c76335dec50d2c953590fb02d259/src/core/reactor.ts#L2342) → [`_next()`](https://github.com/lf-lang/reactor-ts/blob/554a86128c35c76335dec50d2c953590fb02d259/src/core/reactor.ts#L2011) → `_react()`
[...]

PR #217, opened 2023-08-18
Draft: Adjusting access to utils for federated execution

byeong-gil:
Fixes https://github.com/lf-lang/reactor-ts/issues/112. 

This PR is still working on progress. The purpose of the work is to remove functionalities for federated execution from `reactor.ts` and add them to `federation.ts`.


byeong-gil:
@lhstrh If we do not define `UtilityFunctions` anonymously, how can `util` access to `private` or `protected` methods of `App`?

Issue #223, opened 2023-08-28
Port dependency/precedence logic when adding a mutation


axmmisaka:
This is a problem which troubled me for a while and I discovered it might potentially be a rather major issue.

When adding a mutation, it is necessary to declare ports. We declare a port as an argument, as things are set up currently, either as an `IOPort` or as a `WritablePort`. They are, then, handled differently in `recordDeps`: either as dependency or precedence to the mutation.

https://github.com/lf-lang/reactor-ts/blob/f5c195f3b2c276f7f53389b733c3cf4578ca4a89/src/core/reactor.ts#L659-L665

https://github.com/lf-lang/reactor-ts/blob/f5c195f3b2c276f7f53389b733c3cf4578ca4a89/src/core/reactor.ts#L689-L695

The issue is, they are able to be converted to one another inside of the mutation (by using `this.getReactor().writable(port)`, or `(port as unknown as WritablePort<number>).getPort()`), and potentially create a causality loop. Unfortunately to do connection inside of a mutation, it appears that `IOPort` version is required, and if we want to connect AND write to a port then conversion is necessary (either pass in an `WritablePort` and convert to `IOPort` when connecting, or pass in an `IOPort` and convert to `WritablePort`).

axmmisaka:
Here's an example, a fix which makes Prof. Dr. @hokeun 's Online Facility Location benchmark working:

https://github.com/lf-lang/reactor-ts/commit/4a230b6bd93c5d38ed611fcc699d7bc4175e5424

(See diff for more info)

The corresponding runtime graph are as follow:

Not working
```mermaid
graph
0["facilityLocation.0c4e7357[R0]"]
1["facilityLocation.0c4e7357[M0]"]
2["facilityLocation.0c4e7357[M1]"]
3["facilityLocation.0c4e7357.InPort"]
4["facilityLocation.0c4e7357.OutPort"]
5["facilityLocation.0c4e7357.OutPort"]
6["facilityLocation.0c4e7357.OutPort"]
7["facilityLocation.0c4e7357.OutPort"]
8["facilityLocation.0c4e7357.OutPort"]
9["facilityLocation.0c4e7357.OutPort"]
10["facilityLocation[M0]"]
11["facilityLocation.0c4e7357.866cd7ec.OutPort"]
1 --> 0
10 --> 1
0 --> 2
3 --> 2
2 --> 4
2 --> 5
2 --> 6
2 --> 7
2 --> 8
2 --> 9
[...]

PR #226, opened 2023-09-01
Fix Online Facility Location

axmmisaka:
## Action item
- [ ] Verify that this modification is correct and fits the original intention when Prof. Dr. @hokeun wrote it

## Problem 

Sorry for my delayed reply.

In the original code, in `Quadrant`'s second mutation, `toAccumulator` (an `OutPort`) was the destination of two components:
`toNextAccumulator` of `Accumulator`:
https://github.com/lf-lang/reactor-ts/blob/b81e1ce9214835d8e3197af1f2d4b4c7b9cac927/src/benchmark/FacilityLocation.ts#L625-L629

and the mutation (`[M1]` itself:
https://github.com/lf-lang/reactor-ts/blob/b81e1ce9214835d8e3197af1f2d4b4c7b9cac927/src/benchmark/FacilityLocation.ts#L757-L763

I suppose, effectively, the first connection (which doesn't make much sense in the first place as it was connecting an `OutPort` to an `OutPort`), is to make a logic that "once `toNextAccumulator` is set, we also set `toAccumulator`"

My current ugly fix is simply replicate this logic without making any additional connection by accessing a parent port from children: in `Accumulator`:

https://github.com/lf-lang/reactor-ts/blob/82d6e34885f476befcdb9d30ab7ebb546ff4e8a8/src/benchmark/FacilityLocation.ts#L300-L313

Then set both ports when one needs to be updated:
https://github.com/lf-lang/reactor-ts/blob/82d6e34885f476befcdb9d30ab7ebb546ff4e8a8/src/benchmark/FacilityLocation.ts#L359-L374

I suppose this is not a good way to do it...... So I'm all ears

_Originally posted by @axmmisaka in https://github.com/lf-lang/reactor-ts/issues/223#issuecomment-1701992639_

### DIFF
--- a/src/benchmark/FacilityLocation.ts
+++ b/src/benchmark/FacilityLocation.ts
@@ -293,7 +293,7 @@ export class Accumulator extends Reactor {
 
   toNextAccumulator = new OutPort<ConfirmExitMsg>(this);
 
-  constructor(parent: Reactor) {
+  constructor(parent: Quadrant) {
     super(parent);
 
     this.addReaction(
@@ -308,7 +308,8 @@ export class Accumulator extends Reactor {
         this.fromSecondQuadrant,
         this.fromThirdQuadrant,
         this.fromFourthQuadrant,
-        this.writable(this.toNextAccumulator)
+        this.writable(this.toNextAccumulator),
+        parent.writable(parent.toAccumulator)
       ],
       function (
         this,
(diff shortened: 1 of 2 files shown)


axmmisaka:
This is fixed in https://github.com/lf-lang/reactor-ts/pull/232, closing

PR #232, opened 2023-09-07
Refactor Reactor::{connect, connectMulti, canConnect}

axmmisaka:
0. `canConnect` doesn't experience situations where throwing Error is required.
1. `canConnect` is solely for testing and should not throw.
2. `connect` and `connectMulti` should throw. Simply return result from `canConnect` and throw in functions that calls them.
3. A mutation referencing a port for connection/disconnection should NOT result in any causality change *per se* between the port and the mutation

- [x] Refactor tests
- [x] Add descriptions for the enum members
- [x] Implement `ConnectablePort`
~~Note: this task ended up being a rather insane one - I told @lhstrh on Friday that I will make some slight restructuring to make it more sane, but it turns out that my attempts are rather futile and I will commit the rough and crazy one I made on Friday. The issue is that `ConnectablePort` must still be able to be a valid `Variable` and cannot be a copy/wrapper that could be passed into `connect`. After all, `connect` taking an `IOPort` is too big to be changed as of now. So my approach is to make a simple wrapper to bypass the causality interface update. I think a more radical solution (such as making `connect` only take `ConnectablePort`, and unwrap/get the `IOPort` inside of it) deserves another branch/PR.~~
- [x] Fix connection bugs (which is why jest is not passing)
    - [x] Do not automatically fail direct feedthrough if it does not create any causality issue 
    - [ ] ~~Do not limit the scope on local check if causality interface has changed~~ Always do global check
- [x] Discuss with @lhstrh if in-mutation casting should be allowed

### DIFF
--- a/__tests__/InvalidMutations.ts
+++ b/__tests__/InvalidMutations.ts
@@ -40,23 +40,23 @@ class R1 extends Reactor {
       function (this, __in1, __in2, __out1, __out2) {
         test("expect error on creating creating direct feed through", () => {
           expect(() => {
-            this.connect(__in2, __out2);
+            this.connect(__in2.asConnectable(), __out2.asConnectable());
           }).toThrowError("New connection introduces direct feed through.");
         });
         test("expect error when creating connection outside container", () => {
           expect(() => {
-            this.connect(__out2, __in2);
+            this.connect(__out2.asConnectable(), __in2.asConnectable());
           }).toThrowError("New connection is outside of container.");
         });
         const R2 = new R1(this.getReactor());
         test("expect error on mutation creating race condition on an output port", () => {
           expect(() => {
-            this.connect(R2.out1, __out1);
+            this.connect(R2.out1.asConnectable(), __out1.asConnectable());
           }).toThrowError("Destination port is already occupied.");
         });
         test("expect error on spawning and creating loop within a reactor", () => {
           expect(() => {
-            this.connect(R2.out1, R2.in1);
+            this.connect(R2.out1.asConnectable(), R2.in1.asConnectable());
           }).toThrowError("New connection introduces cycle.");
         });
       }
--- a/__tests__/InvalidMutations.ts
+++ b/__tests__/InvalidMutations.ts
@@ -38,11 +38,6 @@ class R1 extends Reactor {
       [this.in1],
       [this.in1, this.in2, this.out1, this.out2],
       function (this, __in1, __in2, __out1, __out2) {
-        test("expect error on creating creating direct feed through", () => {
-          expect(() => {
-            this.connect(__in2.asConnectable(), __out2.asConnectable());
-          }).toThrowError("New connection introduces direct feed through.");
-        });
         test("expect error when creating connection outside container", () => {
           expect(() => {
             this.connect(__out2.asConnectable(), __in2.asConnectable());
--- a/src/core/reactor.ts
+++ b/src/core/reactor.ts
@@ -42,7 +42,8 @@ import {
   Startup,
   Shutdown,
   WritableMultiPort,
-  Dummy
+  Dummy,
+  ConnectablePort
 } from "./internal";
 import {v4 as uuidv4} from "uuid";
 import {Bank} from "./bank";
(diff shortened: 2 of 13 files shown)


src/core/reactor.ts
```diff
@@ -61,6 +61,51 @@ export interface Call<A, R> extends Write<A>, Read<R> {
...
+  DESTINATION_OCCUPIED,
+  DOWNSTREAM_WRITE_CONFLICT,
+  NOT_IN_SCOPE,
+  RT_CONNECTION_OUTSIDE_CONTAINER,
+  RT_DIRECT_FEED_THROUGH,
+  RT_CYCLE,
+  MUTATION_CAUSALITY_LOOP
+}
+
+const CanConnectResultReasonMap = new Map<CanConnectResult, string>([
```
lhstrh, line 76: Should we maybe use String enums instead? https://www.typescriptlang.org/docs/handbook/enums.html

axmmisaka:
After discussion today with @lhstrh we discovered that the current logic of detecting causality interface/direct feedthrough is a bit problematic. I will keep as-is as of now (i.e. no change on enum type/return type) and add tests to make sure that the causality interface issue is resolved first.

axmmisaka:
@lhstrh : 
For reference, the decision regarding code generator:
If we want to connect a port in one microstep as a source, and write to it (or as a destination, but the current implementation wouldn't allow us to read from it I'm afraid), we have two approaches to do this, and they will have the same effect on the causality interface:

Or, should we NOT allow this do be done at all in one tag? This will be logically hard to achieve, I'm afraid, as this way we will need to record if a port is newly connected in the current tag.

1. Declare two `Variables`:
```ts
this.addMutation(
[this.trig],
[this.writable(this.myport), this.myport.asConnectable(), this.destport.asConnectable()],
function (this, wport, cport, dest) {
this.connect(cport, dest);
wport.set(123);
}
);
```
2. Declare one `Variable` and cast inside:
```ts
```ts
this.addMutation(
[this.trig],
[this.writable(this.myport), this.destport.asConnectable()],
function (this, port, dest) {
this.connect(dest, cport);
port.getPort().asConnectable().set(123);
}
);
```
```

File: __tests__/InvalidMutations.ts [hunk 2]
```diff
@@ -38,25 +38,20 @@ class R1 extends Reactor {
...
       function (this, __in1, __in2, __out1, __out2) {
-        test("expect error on creating creating direct feed through", () => {
-          expect(() => {
-            this.connect(__in2, __out2);
-          }).toThrowError("New connection introduces direct feed through.");
-        });
         test("expect error when creating connection outside container", () => {
           expect(() => {
-            this.connect(__out2, __in2);
+            this.connect(__out2.asConnectable(), __in2.asConnectable());
```
lhstrh, line 43: If any port can be converted to a connectable one, then what is the purpose? Moreover, the "kind" of port should be specified in the scope of the constructor, when the reaction is added and its arguments are specified...

lhstrh:
One problem with the diff is that all of the logging stuff is in it too. I think `mutation/full` should have `master` merged into it so that the diff shown in this PR is smaller.

__tests__/InvalidMutations.ts
axmmisaka: This piece of code is more like a temporary short-hand fix.
The purpose of `connectablePort` is not a safeguard for the users, but rather a convenience for the argument processor. If a normal/`writablePort` is passed in, it will add dependencies against the mutation itself. `connectablePort` wont.
Inside the body, if we were to write typescript, this type of casting is inevitable.

axmmisaka:
> One problem with the diff is that all of the logging stuff is in it too. I think `mutation/full` should have `master` merged into it so that the diff shown in this PR is smaller.

Fixed! Sorry

__tests__/InvalidMutations.ts
```diff
@@ -38,25 +38,20 @@ class R1 extends Reactor {
...
           }).toThrowError("New connection is outside of container.");
         });
         const R2 = new R1(this.getReactor());
         test("expect error on mutation creating race condition on an output port", () => {
           expect(() => {
-            this.connect(R2.out1, __out1);
+            this.connect(R2.out1.asConnectable(), __out1.asConnectable());
```
lhstrh, line 49: If a port is made connectable in the body of a reaction even though it was passed into the reaction as writable, we break the mutual exclusion between writable and connectable ports. Probably better to use the same way of obtaining connectable ports as we use for writable ports...

src/core/reactor.ts
```diff
@@ -1148,23 +1199,21 @@ export abstract class Reactor extends Component {
...
-        throw Error("New connection introduces cycle.");
-      } else if (hasDirectFeedThrough) {
-        throw Error("New connection introduces direct feed through.");
-      }
+      if (
+        src instanceof InPort &&
+        dst instanceof OutPort &&
+        dst.getContainer() === src.getContainer()
+      ) {
+        return CanConnectResult.RT_DIRECT_FEED_THROUGH;
```
lhstrh, line 1213: Delete this commented out code.

src/core/reactor.ts
```diff
@@ -42,7 +42,8 @@ import {
   Startup,
   Shutdown,
   WritableMultiPort,
-  Dummy
+  Dummy,
```
lhstrh, line 45: ?

__tests__/InvalidMutations.ts
lhstrh: Will resolve this later.

lhstrh:
While there may be some rough edges in the PR, it's generally a solid improvement, and the tests are running fine, so we should probably just go ahead and merge this...

lhstrh (approved):

PR #237, opened 2023-10-03
Use of `pinojs` as logger instead of `ulog`

### LINKED ISSUE #228: Stop using ULog, migrate to loglevel or winston
ULog is last maintained 3 years ago, has rather poor documentation, very few people use it, and it doesn't have type info published on NPM. We should consider switching to a more modern and well-maintained log library, such as loglevel or winston.

Lucky for us we already have a logger wrapper, so phasing ULog out shouldn't take too much effort.

axmmisaka:
ULog is too old, and while having a log wrapper is fine imo, the current way which we structure it is a bit too bad.
This closes #228 

~~Last TODO: prettify the output so it's not a miserable json~~


lhstrh (approved):

axmmisaka:
Merge https://github.com/lf-lang/reactor-ts/pull/239 first.
Then, https://github.com/lf-lang/lingua-franca/pull/2052 should pass. Merge that.
Finally, drop 039c0e8 and potentially 9402a47, and merge this PR.

lhstrh (approved):
LGTM!

PR #260, opened 2023-12-11
Remove PTAG and ABS messages

byeong-gil:
This PR aims to align reactor-ts with [PR#2118](https://github.com/lf-lang/lingua-franca/pull/2118) in lingua-franca that removes unnecessary PTAG and ABS messages.


edwardalee (approved):
LGTM!

PR #298, opened 2024-08-21
Retrieval of reactor-ts via git and not the local filesystem

byeonggiljun:


lhstrh:
Looks like a good solution to me!

lhstrh:
That makes me wonder though; how does this work in the `lingua-franca` repo? The same fix wouldn't work there.

lhstrh:
On `lingua-franca#master` we use the default `package.json` which currently points to `git://github.com/lf-lang/reactor-ts.git#master`. So we are testing against the correct runtime, it seems.

byeonggiljun:
> On lingua-franca#master we use the default package.json which currently points to git://github.com/lf-lang/reactor-ts.git#master. So we are testing against the correct runtime, it seems.

I also think so. :)

lhstrh (approved):
:+1:

PR #302, opened 2024-08-23
Graceful termination of socket connections to the RTI

byeonggiljun:


lhstrh (commented):
Have you been able to confirm that this addresses the non-deterministic failure we've been seeing?

byeonggiljun:
@lhstrh As a federate destroys the socket connection right after sending LTC and Resign signals, the RTI potentially misses those signals. Thus, I propose using `socket.end()` instead of `socket.destroy()` so that the buffer is flushed before terminating. However, as both `socket.write()` and `socket.end()` are asynchronous, I guess `socket.end()` may close the connection before `socket.write()` writes any message to the buffer. 
https://github.com/lf-lang/reactor-ts/blob/d0286536ae153d8e002ab096eee1eb3daefcb02a/src/core/federation.ts#L1449-L1454

To completely mitigate this, we have to manage a buffer ourselves to track pending network messages that need to be sent. But this change makes it much better than `master` and works fine in most cases.

lhstrh:
Cool! It might make sense to document your analysis in an issue, should the problem resurface.

lhstrh (approved):

byeonggiljun:
> Cool! It might make sense to document your analysis in an issue, should the problem resurface.

I created the issue https://github.com/lf-lang/reactor-ts/issues/303!

